PR #80, opened 2015-01-16
This version can handle upgrading of Flyspray.

jahto:
It's just here waiting in case the project chooses to distribute it's own version. In that case, the upgrader too needs some additional checks and warnings.

### DIFF
--- a/setup/upgrade.php
+++ b/setup/upgrade.php
@@ -270,6 +270,23 @@ class ConfUpdater
     }
 }
 
+function postgresql_adodb() {
+    if (class_exists('ReflectionClass')) {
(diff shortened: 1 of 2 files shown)


jahto:
Forgot to mention explicitly: do not pull yet, just keep the request open instead. This might not be needed at all if I get a refined version of the patch into ADOdb and its next version gets released before Flyspray 1.0. But having the pull request publicly visible might instruct those who really need it NOW to the right github repository, fork and branch. And it's always good to have one more ace in the sleeve.

For those who really need this immediately, after applying the patch: copy setup/upgrade/1.0/datadict-postgres.inc.php to vendor/adodb/adodb-php/datadict/.

PR #105, opened 2015-02-25
the anti csrf and less cookies merge

peterdd:
Maybe a handful of actions still must be makeover for anti csrf token.

Please don't be too upset if some buttons now look a bit ugly, but that was needed to replace simple links by a form. 

This can all be fixed by CSS.

Due some cookies are dropped in favor of urlparams for handling parallel different projects in different browser tabs, the are still some usability issues like the jump to default project when you hit the flyspray icon/project heading.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,9 +2,9 @@
 
 www.flyspray.org
 
-Flyspray Bug Tracking System
+# Flyspray Bug Tracking System
 
-Flyspray is an uncomplicated, web-based bug tracking system for assisting with software development. Learn more...
+Flyspray is an uncomplicated, web-based bug tracking system.
 If you already know all about Flyspray, why wait? Download it now!
 If Flyspray is helping your business, please consider helping us by donating a couple of dollars. 
 Be added to our generous donators page today!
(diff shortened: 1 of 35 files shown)


Psychokiller1888:
How would you rate Flyspray security after your work is done? That's a question I'm often asked...

peterdd:
That was only one step in the right direction I think. 

For the rest I cannot say much at the moment as I'm not a professional penetration tester or something like that and did not review the whole code in that direction.

I just stumbled over the csrf token issue in another software which is used often in Germany. Now I check my and at work used software if they know and address these topic.

What I know is that the integration of ckeditor (or its predecessor fckeditor) also needs attention when integrated into a project.
There was a security break on a project due to untaken unsecured/missing authentication file manager feature in the editor, so attackers could upload malware to server without authentication.
(That I had in mind when I saw that flyspray's ckeditor integration with many plugins is >400KB in size)

Well, I haven't dig into the current integration of ckeditor in flyspray yet, so I don't know if it is done right or wrong.

And then there is the world after snowden with its daily apocalypse IT news like the SSL mess, mobile sim card keys stolen at the world leading manufacturer,  and so on...

PR #122, opened 2015-03-03
Translation tool in admin area

peterdd:
@Psychokiller1888 :my android tablet to the rescue for sendin pull req.,lol


Psychokiller1888:
@peterdd Display seems broken:

http://postimg.org/image/ofs4yrpk9/

Emptied cache and all.. I'm on Chrome, Win 10

peterdd:
It is not broken, it is just not yet full integrated as I wrote on commit. :-)

I just copied the two files to scripts/ and adapted until it roughly works, the HTML is surely invalid because translation sends output before the page template renders. :-)

Just for us devs at the moment, so we can use that easier translation workflow immediatly. I know the work with template for translation tool must be done. And naturally that tpl_form thing too..

If you want you can work on full valid integration into admin area further, I have no more time today on that, sorry.

So its not a problem of your browser, it's because of my laziness.

Psychokiller1888:
Oki cool, no problem, just thought that went unseen. Well, can't right now, i'm at work :) But we have time

danoh:
Did you tried to use gettext for the translation? I prefer gettext to any other solution, while we have editors for gettext, which can check spelling, translation sugestions and so on. Gettext can also handle plural forms.

peterdd:
I like that it is kept so simple in flyspray just by using the flyspray itself in browser and everyone can contribute without searching what gettext is, how it works and additional software installs. I remember fiddling with .po files, but never got the advantage in opposition to just an array of strings everybody can look at with dead simple editors. Keep simple things simple. Just my 2 cent.

Psychokiller1888:
Agree

PR #125, opened 2015-03-03
Fix 'undefined index' for toplevel view

jahto:
When viewing 'all projects' and project = 0.


jahto:
This is unfortunately somewhat a workaround of having the data needed not available otherwise. For 1.1, I'd propose another solution. See includes/class.flyspray.php line 387. Why not have all the project data available there, make sure it really gets updated whenever a projects settings changes (probably already happens), do something to that FIXME (I've already fixed a few places where deleted projects broke the assumption that the array is indexed py project id...), convert it into an array of arrays indexed by project id, and just fix the errors resulting from this. Too much work for 1.0.

PR #142, opened 2015-03-06
Update sv_SE.php

thinegod:
Fixed typo


Psychokiller1888:
Thank you, merged, but sv_SE is still missing over 200 strings...

jahto:
I think one co-worker of mine could help with that, although what he speaks is actually sv_FI, what our swedish minority in Finland speaks, but they do handle sv_SE also quite well. I might myself add finnish sometimes later in beta phase, not a necessity, almost everyone here can handle english, and surprisingly, could also help with getting hungarian translation entries up-to-date. Although some real hungarians will complain a little...

Psychokiller1888:
If you want I might have someone in Sweden. As well as an hungarian guy and a Russian one. I'll let FI to you

jahto:
Fine with me. So I won't do anything unless those persons just do not have the time. Quite comprehensible translation is better that no translation at all.

Psychokiller1888:
I'll check and ask and will give feedback soon

jahto:
Not all translation entries need to be perfect in beta phase. It's fine, in my opinion, to get the missing ones before final 1.0. They are not new features. And we'll anyway will be missing a few ones even at that stage. I, for example, am not very good at translating the chinese entries. It's all chinese for me.

PR #147, opened 2015-03-07
permissions on myprofile, reload userinfo on usersetting safe, ..

peterdd:
and dont let users do virtual flyspray suicide


Psychokiller1888:
Dude, just tested now, it's awesome!

PR #154, opened 2015-03-09
Fixed install once again.

jahto:
I didn't check if rightmost values in group permissions are actually right ones, just removed one extra value.


Psychokiller1888:
Yep, weird, just made sure and your fix is correct. Have to check what that extra value was, or if it was me miscounting last time

PR #166, opened 2015-03-11
Once again, fix install and upgrade process...

jahto:


Psychokiller1888:
Damnit! I start to hate that installer!

PR #174, opened 2015-03-13
View task permissions

jahto:
Already somewhat tested. It's coming in a bit late, but both I and Jordan have a real use case for it, so think really carefully before rejecting...

### DIFF
--- a/includes/class.flyspray.php
+++ b/includes/class.flyspray.php
@@ -1229,4 +1229,13 @@ class Flyspray
         return '';
     }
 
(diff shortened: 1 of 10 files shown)


Psychokiller1888:
Why on earth would I reject it??

jahto:
Yeah, absolutely no reason. Thanks anyway! And now I'll enjoy a really well-deserved beer.

Psychokiller1888:
Enjoy your evening! Olut perkele!

PR #180, opened 2015-03-14
Few things discussed for 1.0

Psychokiller1888:
Such as url_rewriting in admin
Default order direction
FR missing translation
FR installer translation


jahto:
Just a reminder: the string "off" evaluates to TRUE in php, so `if ($fs->prefs['url_rewriting'])` doesn't work as expected and my test site just broke down. I think it's time for a beer...

Psychokiller1888:
Woot? I have been testing this back and forth, so what broke?

Psychokiller1888:
#181

PR #182, opened 2015-03-14
Get rid of a few remaining "missingtoken" messages.

jahto:
There's probably some more still lurking around somewhere...

### DIFF
--- a/includes/modify.inc.php
+++ b/includes/modify.inc.php
@@ -2077,10 +2077,10 @@ switch ($action = Req::val('action'))
 
         //set the subtask supertask_id to 0 removing parent child relationship
         $db->Query("UPDATE {tasks} SET supertask_id=0 WHERE task_id = ?",
-                   array(Get::val('subtaskid')));
+                   array(Post::val('subtaskid')));
 
         //write event log
-        Flyspray::logEvent(Get::val('task_id'), 33, Get::val('subtaskid'));
+        Flyspray::logEvent(Get::val('task_id'), 33, Post::val('subtaskid'));
         //post success message to the user
         $_SESSION['SUCCESS'] = L('subtaskremovedmsg');
         //redirect the user back to the right task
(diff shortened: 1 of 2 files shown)


jahto:
Either this or something else caused something very strange to happen afterwards. Investigating... Could be that this one must be reverted.

Psychokiller1888:
What strange happened? I'm not on my computer,  but tell me if needed

jahto:
Oops. Saved the source code of my test case page in flyspray root directory with the name "index.html" to inspect it in netbeans and check that the source seems ok, forgot to delete it, and what did I get back everytime... Problem solved.

peterdd:
Thx. Let's bet how many location we still find and fix until release 1.0 . :-P 

My bet is 2.

(Haven't reviewed for that this week so it is a fair bet. :-)  only existing dev code of today)

jahto:
I'm not making any bets yet, there's still tomorrow left to find them. Hint: search the code for `<a href`.

Psychokiller1888:
There's still more time, tomorrow it's no new feature, we still have plenty of time for debug... Will pack alpha rev on monday i guess, but still, alpha means alpha, bugs can be found and will still be found until final release.

peterdd:
I vote for notification icons in fs1.0 if new events or notifications exists. I dont read emails from automated services like facebook etc, but like them when i'm on the website.

Psychokiller1888:
That means storing notifications internally for future reading as well... FS 1.1 feature maybe? :)

jahto:
I might be able to cook up something in the few remaining hours before freeze. If peterdd is satisfied with a very simple version. Combining what already exists in class.notify.php and events.inc.php it might be possible to add a notification that something just happened, in the same way as it shows in a tasks history. Then, show an icon if there are unread notifications and add a way to mark them read.

No use to waste time with discussing about this more. I'll just open a new branch again and test if it's possible.

jahto:
Peterdd, can you implement the user interface really fast, ie. before midnight, while I'm still working with backend things? I've got now something implemented on my branch online_notifications. Most interesting one is static function Notifications::GetUnreadNotifications, returns an array of message_ids and message_bodies. I need something to the user interface that posts an array of ints (message_ids) when something has been marked as read (they will be deleted). Let's call the action notifications.remove and I'll add it to modify.inc.php.

The messages still need different formatting, email and jabber formats do not fit a user interface.

jahto:
Got it working, borrowed some things from jabber, which stores the notifications and sends them later. Removing already seen ones still to do.

jahto:
@peterdd, you have 8 hours remaining to implement the user interface if you want online notifications if FS 1.0. Backend has been done, but needs of course still a bit refining.

jahto:
@Psychokiller1888, what should I do with this one feature? The backend has already been done. User interface missing. To make a pull request or not to make a pull request?

Psychokiller1888:
Make a pull request, If alpha comes out tomorrow or Tuesday it doesn't change a damn, FS 1.0 is awaited since over 2 years, so a day or two more won't kill anybody anymore

PR #187, opened 2015-03-15
Online notifications

jahto:
Backend work done. Didn't immediately break everything. Further testing needs a user interface. Peter, hurry up! It's you who voted for this feature to be in 1.0.

### DIFF
--- a/includes/class.flyspray.php
+++ b/includes/class.flyspray.php
@@ -1205,6 +1205,7 @@ class Flyspray
         $return = array(0             => L('none'),
                         NOTIFY_EMAIL  => L('email'),
                         NOTIFY_JABBER => L('jabber'),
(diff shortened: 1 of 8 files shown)


jahto:
I suggest that we release anyway on 16. even if the user interface is not done, unless I really broke something. Hope not, I've been making modifications already and the notifications itself are already made.

Psychokiller1888:
Peter is not reading notification email, so that's why he doesn't see :) :) :)

jahto:
He's probably too busy trying to implement his favorite feature for 1.0 to notice that it has already been implemented. Or drinking too much because he believes it will not be in 1.0.

peterdd:
No, today sleeping, meeting family, regeneration.

PR #190, opened 2015-03-15
Guess what... fixed installer.

jahto:
Upgrade process from 0.9.9.7 works, although something must be done those messages.


jahto:
Does this ever stop? I need now a beer and my favorite hungarian spirit, Unicum. PLEASE test creating a new database and a new Flyspray installation on it before committing or making pull requests to anything that touches the database schema or values inserted. Although have to admit, I too have broken it sometimes... Being more careful now.

Psychokiller1888:
You know what? I just don't understand, because before my last pull I actually upgraded And removed FS for fresh install. So I just don't get it and was very surprised seeing that

jahto:
Could be that mysql is more forgiving than postgresql and tolerates one final missing quote. Have had already a few experiences with that kind of differences. Was something related to using a field in GROUP BY and not specifying an aggregate function using that same field in SELECT clause. SQL standard (2003 version or something like it) requires it, mysql just makes a good guess what the user might have been meaning if it's not used. And again, or something like it.

By the way, no pull requests coming from me today anymore, unless absolutely necessary to fix something I didn't notice. Will check the situation tomorrow morning. Happy upgrade night!

Psychokiller1888:
Maybe. Once again sorry.

Ok about your work, thanks for it. I'm checking at url_rewriting a little still and will also stop. I'll write in the morning then

PR #194, opened 2015-03-16
Really get rid of undefined offset 2? And a few more small ones.

jahto:


Psychokiller1888:
@peterdd @jahto do you guys want to finish this notification stuffs before I pack alpha? Would be nice...

jahto:
I hope it was the right fix for registration too. Would you test a little and check if you find some more broken things. Registration itself and adding multiple tasks now works for me. There might still be small problem with pending new user registrations that must be approved, the table is empty for me.

And notifications MUST be finished at least so much that they don't break anything. The rest is upon peterdd, if he want's the user interface too into alpha.

jahto:
Ok, tracked down that missing pending user registration. Invalid SMTP server settings, message got not send, and so user pending user registration was not added. Nothing to fix (?).

Psychokiller1888:
Yeah, I went through all registration sorts but oauth as it won't work locally. Seems ok. Now migrating fs logo and resized it, as it has nothing to do there and the logo should be theme based as well

PR #197, opened 2015-03-16
Can't be sure if effort has been fetched or not.

jahto:


Psychokiller1888:
Will add later, you're not up to date, can't merge. First fixing everything i find, and there's things to do...

jahto:
Now I should be up-to-date. Note that I'm unavailable between 5 and 9 CET today, so can't fix anything on the evening.

Psychokiller1888:
Yep, no problem, take it easy. Waiting on peterdd opinion about notification, but as you said, it must be in alpha, so i'll wait on UI implementation

PR #204, opened 2015-03-20
At least honour the setting...

jahto:
This does not solve FS#1954 but at least what is sent is sent according to the setting. Could someone test? I'm not currently able to send email using SMTP, can only use web interfaces...


peterdd:
will send my pull request too, but probably collides with your merge.

jahto:
Merge my commit first and fix the conflicts if you've got a better solution. I can't test just now.

PR #205, opened 2015-03-20
send plain and html mail together

peterdd:


peterdd:
close conflicting pull request, done now with direct change to master.

PR #207, opened 2015-03-20
add ldap patch

peterdd:
based on provided patch for 0.9.9.7 on bugs.flyspray.org, modified for 1.0.
Untested, works only as stub for later dev.

### DIFF
--- a/includes/class.flyspray.php
+++ b/includes/class.flyspray.php
@@ -630,6 +630,7 @@ class Flyspray
      * Check if a user provided the right credentials
      * @param string $username
      * @param string $password
+     * @param string $method '', 'oauth', 'ldap', 'native'
      * @access public static
      * @return integer user_id on success, 0 if account or user is disabled, -1 if password is wrong
      * @version 1.0
@@ -660,6 +661,7 @@ class Flyspray
             return 0;
         }
 
+        if( method != 'ldap' ){
         //encrypt the password with the method used in the db
         switch (strlen($auth_details['user_pass'])) {
             case 40:
@@ -672,7 +674,7 @@ class Flyspray
                 $password = crypt($password, $auth_details['user_pass']); //using the salt from db
                 break;
         }
-
+        }
         if ($auth_details['lock_until'] > 0 && $auth_details['lock_until'] < time()) {
             $db->Query('UPDATE {users} SET lock_until = 0, account_enabled = 1, login_attempts = 0
                            WHERE user_id = ?', array($auth_details['user_id']));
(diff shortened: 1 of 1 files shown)


peterdd:
It should not break anything even on feature freeze.
Its more to have a starting point for the dudes who need it. (i don't)

Psychokiller1888:
Adding LDAP support can lead to severe security issue which in my opinion, tell me if I'm wrong, aren't able to test until next week or in two weeks beta that should be data security secured

peterdd:
It will be in no way activated for 1.0.
And this patch just add the function stub, but can't be used accidently with this patch.

Maybe add some comments in the code to make that clear.

I have no experience with LDAP, and if that adds security issues it is currently up to the people who want and need LDAP within flyspray. And this roughly made patch seemed to be used by archlinux users, as they also provided packets with this patch.

For all the others it is justs dead code yet.

Psychokiller1888:
It's dead code, doesn't seem to be able to brake other pw checks, so I'll merge it, but no more commits about this before beta is branched... Please open a task on FS, this will need serious work, as nothing is made, such as the port that is unusable for a synology LDAP directory per exemple

peterdd:
yes, same opinion. :-)

jahto:
Ok, code that is dead can be merged. This is something I too looked at, patches to do it have been flowing around and already half-way tested by users. A good candidate to be enabled already in the inevitable 1.0.1 after some testing. I've got already some code for that too... related to notifications.

peterdd:
BTW and I think it is important too for communication external, interested peoples: Flyspray Version numbers!

My opinion is we should go the way of 1.0, 1.1, 1.2, 1.3 for this year.
Releasing one .1 number planned every 3 months or so - but no hard fixed dates, no hurry - feature freeze times ok and the fixing only until release, but release only "it's done when it's done".

We do constantly fixing things, adding normal to small things. We will not change very fundamental things, we do not do a rewrite.

A 2.0 means a complete rewrite or architecture change. I don't see a good reason for that.
A 1.0.1 is a tiny patch which adds no new feature, only quick fix for a bug.

jahto:
Not exactly of the same opinion. 1.0.1/2/3/etc for fixes/features that don't need any database schema changes or struggling with composer again and can be applied by some users even by just doing `git fetch upstream master; git merge upstream/master`. 1.1 for changes that need running upgrade process and maybe something else too.

On the move, accu emptying, no power available nearby, have to logout from the discussion for awhile.

peterdd:
Ok. This week we got aware of too many yet unknown issues of flyspray. (making it urlrewrite ready again, notification and events, experimenting composer install with installer, mail bugs.)

I wish you a nice weekend of recreation. A weekend without electronic devices, just nature, friends and beer?

PR #214, opened 2015-03-23
Make accepting re-open request work.

jahto:
Hope this does not require url-rewriting work also. And I really should learn some day how to use styling and make those buttons look reasonable.


peterdd:
an Easy way is to use the builtin tools of browsers or firebug plugin. You can fiddle around in the live page and see in the tools how all the rules tag-css, class-css, id-css, inline-style-css and the hierarchy and rule combination work together.

PR #215, opened 2015-03-23
#Discuss it - probably unneeded exit in index.php

danoh:
If user sends to someone other URL from the browser, this exit prevets the user to use the URL and login to Flyspray, because only white page is displayed.


peterdd:
I think this must be discussed with reviewing Flyspray::show_error and exception handling together.

The hard exit calls are probably made as last defense line for a reason like unwanted information leakage. 

Your intention seems absolute legit, sending a link to a project and then enable to login should be possible. On the other side, info about the project for visitors (before login) may be forbidden, even the project title or the language of the project.

danoh:
Just for info, if you send link to task, you get login box, if you send link to project, you are banned. It is a little bit inconsistent.

peterdd:
Ok, could you add that to bugs.flyspray.org please, so we have it on our todo list?

PR #225, opened 2015-03-26
submit button for the jsless peoples

peterdd:
The second button for the task search is optional even for the jsless peoples, but made them visible for them.
Think of touch displays...
Hope psycho can live with that.

Without the buttons for jsenabled users the select and quick task search fields now looks like one form, but they are completely different forms ...


peterdd:
The user can't know at first that the project selector hasn't anything to do with the quick task id search.

Maybe he thinks "Let's look for #1 in project 1", then "#1 in project 2".. (false assumption)

With two submit buttons the user may think, "oh the project selector is one thing", and the "taskid search is another separate thing". (currently correct assumption)

But all that must be seen in bigger context. This is only a tiny issue.

PR #227, opened 2015-03-27
Update composerit.php

peterdd:
psycho: I don't understand what was wrong with my patches.. You can 'require composerinstaller', but not 'composer.phar'. I tried before, and if it is possible to do it in fewer steps I would do.

And I prefer indentation only if it makes sense.
Indentation html output after ```<html> and <body>``` doesn't help in reading the output code and costs 2 Bytes per output line. 

Similiar with indentation of mixed php with html, IMHO.

### DIFF
--- a/setup/composerit.php
+++ b/setup/composerit.php
@@ -35,10 +35,9 @@ header("Pragma: no-cache");
 <?php
 #chdir('..');
 # This works also in php5.3.* with SAFE_MODE enabled
-if (ini_get('safe_mode') == 1) {
-	$composerfile = file_get_contents('https://getcomposer.org/installer');
-	file_put_contents('composerinstaller', $composerfile);
-}
+$composerfile = file_get_contents('https://getcomposer.org/installer');
+file_put_contents('composerinstaller', $composerfile);
+
 if (!is_readable('composerinstaller')) {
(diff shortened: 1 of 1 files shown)


Psychokiller1888:
Refused, sorry. If I changed it, it's because you stuffs wasn't working and that's it. I made it work for me. Getting rid of that damn composer install output is a must, as it returns me errors where everything works fine. That's why the debug bool.

peterdd:
Please reread my patch. First it just downloads a file - simple for every case.
Later do the require of composerinstaller (not composer.phar!) 
This "require composerinstaller;" creates the composer.phar!

What you do is shell_exec() and piping the composerinstaller download direct into a new php process, which is impossible here. So I made that if SAFE_MODE switch.

My patch was splitting this 2inone into 2 separate steps, which is also easier to debug. (you can see if the file composerinstaller is downloaded)

peterdd:
can u show me what wasn't working?

Psychokiller1888:
What more to say? Failed to load cacert.pem, where it's not needed, it's included in composer.phar. Failed to download, composer on line 24 fatal error, no vendors downloaded, nothing. And of course, no user return

PR #246, opened 2015-04-01
Memory fix

peterdd:
This is a simple fix for the task list memory problem when many tasks exists in a database.


Psychokiller1888:
What editor are you using for php?

peterdd:
Nano on ssh on server and editor on github.com for commits.

Psychokiller1888:
Just saying, cause my editor will always fix all these:

if($foo==$bar) => if ($foo == $bar)

upon ctrl+s, to standards, so it makes useless changes loosing the focus on what i actually did. Can't you set up a syntax auto checker?

peterdd:
Mmh, I have color coding for php on nano working, but it is justs based on some regular expressions. No real php validation or coding style enforcing.

So I have to add this additional spaces of if/foreach/while.. in future?

Psychokiller1888:
Well the thing is that it analyses the document and defines the current styling. The problem is that sometimes you have the spacing, sometimes not. In my opinion, better to space all the time after any condition.

Then again, as said, I'll run the parser on FS 1.0 for a completly rewritten 1.1, same convention everywhere in every files. So we can discuss it later

peterdd:
ok, will try it in future.

yea, coding style can be discussed a bit, especially on mixed html/php. But yea, later.

Psychokiller1888:
html/php mix is retarded and will be removed for 2.0 that's for sure, let's go mvc

PR #252, opened 2015-04-01
Because i'm borred with that remove setup dir

Psychokiller1888:


jahto:
Yeah, makes me too insane always renaming it to something else before dong testing and then renaming it back to the original before committing. Once even made a commit that deleted the whole dir and had to revert it.

Psychokiller1888:
Yeah, I forgot that i could just simply name that to alpha 2 and include dev in it so it won't bother us with the setup dir anymore...

jahto:
We could also think about releasing that alpha 2... There's nothing really critical on my to-do list to prevent it. No use to release it tomorrow, people will have their easter holidays. But maybe on tuesday. Just to show that this project really makes some progress after a few years and a few days more to check if there's still something that doesn't work.

PR #264, opened 2015-04-06
That was forgotten by me

Psychokiller1888:
I've been asked a few weeks ago to add a config for max votes a day...
Also fixed the logic with >= as the check is made BEFORE voting...


Psychokiller1888:
And I just got reminded about it by the guy, lol...

peterdd:
I think a concept like 'x votes for open tasks' independent of days or any time delay attractive.

Any user has a fix amount of votes available which he can use to make his votes on tasks.
But this needs the ability to 'unvote' a previously made vote for a task.

What is interesting with this concept? The user has to think a bit about which feature/bug is most important for him to vote. And makes users more equal on voting for tasks.

Maybe 5 to 10 for the vote limit is ok? If I remember right uservoice.com uses 10 votes, but there you can make up to 3 points of your 10 points for a task to weight your votes. 

Technically it would be just counting the open tasks the user voted. If the limit of x is reached, just deny any further votes until the user canceled a vote or a task where he voted is closed.

If a closed task on which the user voted before, is reopend, this could lead to a bit overvoting. But that is no problem. The user must just wait until the open tasks he voted drops under the limit or he must adjust his votings.

Psychokiller1888:
Yeah, the thing is that he's using it more like a "Me too!" feature and that that hardcoded 2 was just blocking for him since long. As we are in freeze I did not change anything, just added the possibility to edit without php edit lol. Your idea is good and should be implemented for 1.1 maybe.

PR #280, opened 2015-04-19
New get task list

jahto:
And a few notes:
- Do not blindly merge this assuming that I'm right (although I know I am...).
- The queries produced have been only very lightly tested with mysql too, but seem to be correct.
- I'd really appreciate it if @peterdd and @danoh check this before merging it.
- There are no obvious bugs.
- There must be some inobvious bugs still...
- There's still a lot of debugging and commented-out code that can be removed before going final 1.0.
- Formatting is might not the best one.
- It could be made even faster by using some clever tricks and really tuning indexes, but I've got no time for that anymore, if beta should be released soon. After that, it's really just possible bug fixes.
- I will have a second round thinking this for the next version.
- Anyway, it's the best and fastest solution we have now. Beats any other version and actually survives paging and searching with over 150000 open tasks, 750000 comments and 1500000 attachments, although with that insane and unrealistic mass, it's still a bit slow.


Psychokiller1888:
So, tbh, what's the use of making a pull request if it's to wait? Dude, you made an awesome work, that was even planned, and I gonna merge it and that's it... We will try, and we will debug this week and as planned beta will roll out

PR #301, opened 2015-06-19
avoid the warning "open_basedir restriction in effect"

mouts:
fixes a warning triggered by is_file(). see more [here](https://groups.google.com/forum/#!topic/flyspray/uWg09IuJfUc)

note that every theme directory that start with '.' will be ignored

### DIFF
--- a/includes/class.flyspray.php
+++ b/includes/class.flyspray.php
@@ -412,7 +412,7 @@ class Flyspray
         $dirname = dirname(dirname(__FILE__));
         if ($handle = opendir($dirname . '/themes/')) {
             while (false !== ($file = readdir($handle))) {
-                if ($file != '.' && $file != '..' && is_dir("$dirname/themes/$file") && is_file("$dirname/themes/$file/theme.css")) {
+                if (substr($file, 0, 1)  != '.' && is_dir("$dirname/themes/$file") && is_file("$dirname/themes/$file/theme.css")) {
                     $themeArray[] = $file;
                 }
             }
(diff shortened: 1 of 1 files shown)


includes/class.flyspray.php
```diff
@@ -408,18 +408,19 @@ public static function listProjects(/*$active_only = true*/) // FIXME: $active_o
...
     {
-        $theme_array = array();
-        if ($handle = opendir(dirname(dirname(__FILE__)) . '/themes/')) {
+        $themeArray = array();
+        $dirname = dirname(dirname(__FILE__));
+        if ($handle = opendir($dirname . '/themes/')) {
             while (false !== ($file = readdir($handle))) {
-                if ($file != '.' && $file != '..' && is_file(dirname(dirname(__FILE__)) . "/themes/$file/theme.css")) {
-                    $theme_array[] = $file;
+                if ($file{0}  != '.' && is_dir("$dirname/themes/$file") && is_file("$dirname/themes/$file/theme.css")) {
```
peterdd: I made a pull request too.

Didn't know you can access strings like this in php. But found this on
http://docs.php.net/language.types.string :

```
Strings may also be accessed using braces, as in $str{42}, for the same purpose. However, this syntax is deprecated as of PHP 6. Use square brackets instead.
```

Psychokiller1888: Yeah, finally they get rid of that... Use square brackets is logic, as a string is nothing more than an array of chars

PR #318, opened 2015-07-20
anti csrf token checks for data changing ajax requests

peterdd:
Quite big patch. Please review and test.
Fix for FS#2006


peterdd:
Thx. Could you also please fix FS#1996 (readding ajax_load.gif to CleanFS/ )?
You can use the file from the old blue theme. I don't have the rights to do so.

PR #331, opened 2015-07-24
use $do instead of urlparam for body class

peterdd:
because of twisted logic with project switcher


peterdd:
made another pull request now without merge conflicts..

Psychokiller1888:
Ok thanks, was going into it this evening. Sry, working like 16hours + a day now until end of august, it's high season for us

PR #343, opened 2015-07-27
Native language notifications

jahto:
Ok, this is quite a big one. Most parts have been tested in production since Friday morning, some later additions since Monday morning. Nothing severe detected. Of course there is somewhere some corned case I didn't see, nor was tested. So I'd say, merging is halfway safe. It shouldn't break everything, but maybe something...


includes/class.notify.php
```diff
@@ -18,105 +18,145 @@
...
-      $row = $db->FetchRow($result);
-      $message_id = $row['message_id'];
-
-      // If message could not be inserted for
-      // whatever reason...
-      if (!$message_id) {
-          return false;
-      }
+	// grab notification id
+	$result = $db->Query("SELECT message_id FROM {notification_messages}
```
peterdd: micro optim.: Isn't it possible to get the insert_id from the INSERT directly with adodb?
Avoids (theoretic) race condition when another notification is stored between the INSERT and getting the message_id of it.

And saves one sql request and variable $date not needed.

jahto: I don't know. Have to check adodb docs. But the race condition is not only theoretical. It WILL happen someday somewhere. If it's possible to get the id of an inserted row with adodb, we should check every place where something like that has been used.

Logging out now, probably not available later today. Have to take a shower (gardening), eat and go to sleep.

peterdd: Agreed.

peterdd: this is how it's was changed a year ago for postgres in adodb:

https://github.com/ADOdb/ADOdb/commit/e4563200ccba0982807695d7b4cd2ace4f4e838c

PR #344, opened 2015-07-28
Default 1 wasn't a good idea.

jahto:


peterdd:
Could you please explain a little bit?

jahto:
Happened at workplace after doing upgrade. Every customer here was first assigned to a group "outsiders" and then added to their real project group (about 15 of them currently). Ok "outsiders" now got the permissions to see groups task and projects, so every customer got permission to see almost every project and almost every task (a few are for internal use only, did not affect those projects tasks). Not the wanted result. Happily most everyone in this country is on holiday, so it was not catastrophe. One customer who is working notified me that he's seeing something more than before and I fixed the permissions in our database.

peterdd:
Ok, also all other permissions are set to 0(no perm) by default, so it it consistent too and admins have to tick always the permissions they allow for a new user group.

PR #347, opened 2015-07-28
fix brackets in class.notify.php

peterdd:
push as fast as possible, currently gives php error.

Reminds me of adding an automatic php code check for travis-ci on each commit.

### DIFF
--- a/includes/class.notify.php
+++ b/includes/class.notify.php
@@ -939,7 +939,7 @@ class Notifications {
             }
 
             if (($fs->prefs['user_notify'] == '1' && ($recipient['notify_type'] == NOTIFY_JABBER || $recipient['notify_type'] == NOTIFY_BOTH) ) || $fs->prefs['user_notify'] == '3' || $ignoretype) {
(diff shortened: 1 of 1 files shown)


jahto:
Strange. Why didn't this come up in testing? What did I actually test?

peterdd:
probably you had no jabber recipient :-)

I haven't working jabber yet (didn't received anything yet from flyspray installs but don't know why, didnt digged), but I have an entry in a test user and an account I use for testing for psi+ on jabber-ubuntu.de

jahto:
Correct, I do not currently have any jabber recipients at work. But I did have on my home installation, at least once, and also a xmpp server. Tried to find out what's going with a bug or issue reported sometime in spring, can't find it now. The case was flyspray not reading data back from the connection when using TLS. Might be related to http://php.net/manual/en/migration56.openssl.php, php version used was in fact 5.6. My testing stopped because I didn't have the time to struggle to get it accept a self-signed certificate. If you are using TLS, can you check what happens?

PR #349, opened 2015-07-28
fix sql error when user has not an allowed project

peterdd:


jahto:
Yes, our supported databases do optimize it away. Personally I prefer the alternative WHERE 0 = 1.

peterdd:
yea, a bit easier to read for who don't know much about sql..

PR #354, opened 2015-07-31
Add code to allowed tags.

jahto:
Seems safe. A must for comments.


peterdd:
Only for the ckeditor. Well, i fear enabling this a bit, but maybe I'm wrong here.

I stick to dokuwiki for my projects. :-)

PR #379, opened 2015-08-14
add missing Insert_ID() for the db wrapper

peterdd:
This is IMHO important for the elimination of some race condition in the flyspray code.

Better use the autoincrement feature to avoid that. Instead of precalculating with an extra sql query. When precalculated, a second db connection could insert between the 2 queries leading to wrong data.

### DIFF
--- a/includes/class.database.php
+++ b/includes/class.database.php
@@ -117,6 +117,16 @@ class Database
     }
 
     /**
+     * Insert_ID
+     * 
+     * @access public
+     */
+    public function Insert_ID()
+    {
+        return $this->dblink->Insert_ID();
+    }
+
+    /**
      * CountRows
      * Returns the number of rows in a result
      * @param object $result


jahto:
Have you already the next patch that adds the usage of that missing function ready and waiting? If not, I could go through the code during weekend. Shouldn't take more than a few hours.

peterdd:
Not ready yet, but I'm working on big patch for the tag feature (in peterdd master branch, where i use Insert_ID() one time, 3 more existing candids for Insert_ID() partly detected with comments in code in create_task and list_* stuff.

Will do only pull request when ready and tested so nothing breaks in Flyspray master.
The existing tag feature was like implemented 10% and made no sense, so better change it now than hard refactoring later.

jahto:
I can tell you that there are a lot more than those 3... so I might do at least some testing during the weekend.

jahto:
I must disagree. Gettings tags to work would also mean once again 
changing the database schema. Almost strict no-no-ever-keine-möglichkeit 
after feature-freeze. Although myself changed 2 defaults afterwards. 
Could you think of something like putting those of your changes to a 
future branch, like 1.1. I've also got a few changes waiting for that, 
for example updating dokuwiki, but those changes were not wanted because 
of changing/adding/deleting approximately 3700 files.

But using Insert_ID would not be a new feature. Just fixing a bug. Also, 
like me finally making pull request for native language notifications 
because of your request. The patch was already waiting somewhere there 
in April in my personal archives, just not finished and tested. And it 
was something between feature and bug fix.

Ok, but what was my point, hardly remember myself. I think it was about 
deferring some things to the future.

Mit freundlichem Gruß:

-- Jouni

  08/14/15, 09:23, peterdd schreibt:

peterdd:
The tag feature how it was added in 2013 is a complete misconception, the implementaion looks like an idea after 3 beer, coded 15 minutes after that, then registered it isn't so easy and stopped working on that forever after that evening. *rant*

Keeping it as it is and ship with FS1.0 IMHO would be a mistake, which would require much more effort to rework and smooth upgrading in a FS1.1.

The only alternative I would accept is dropping the existing tag feature for the FS1.0 release.

peterdd:
And here this I promise (you can nail on me for that):

My tag feature refactoring works after completing for FS1.0:
  * from upgrading 0.9.9.7 and previous versions using the upgrader
  * upgrading from github master versions installed (running the upgrader too)

Table 'tags' of current FS1.0 github master keeps untouched, marked as deprecated and be removed in FS1.1. I add new tables for FS1.0 and use mostly existing Flyspray code.

jahto:
Ok, I trust you to get it right. I still somehow feel that tags should just be commented out and deferred until 1.1.

And you are wrong in the assumption "idea after 3 beer, coded 15 minutes after that" would not bring in any results. In fact, it's quite exactly what I do after my daily work. I just don't make those pull requests until I'm sure about the idea. Some of them got dropped before I made a pull request.

But, using Insert_ID definitely should go into 1.0 and that's something I'm capable of getting right even after 15 beers. Unless I'm already sleeping.

jahto:
Ok, gone through the code and hopefully found most of the places where insert_id should be used. Minimal testing, seems to work. You can look at my branch insert_id and check if it works with MySQL too. Don't have time just now for more testing, I should already be at garden, the vegetables need some water.

PR #381, opened 2015-08-16
Upgrader was again broken...

jahto:
And this time, fix it the right way, so we can add the SQL statements in upgrade.xml instead of upgrade.php.


Psychokiller1888:
Wasn't me this time :)

jahto:
Yeah, it was Peter... But now it's fixed the right way and at least that bug won't happen anymore.

PR #387, opened 2015-08-20
fr lang update

supersquared:
- Some words was not appropriated.
- Fixed typos and grammar.
- Fixed accents.
- Added missing translation


jahto:
Just have to comment on this. I was initially a bit frightened when I saw that pull request #385. Because of seeing Fred/supersquared's comment before: 'The only thing is the poor french translation.'

Not actually understanding french, I don't think it was poor, because the translations were made by a native swiss french. But just maybe having a strange dialect that's not standard french. Seems you anyway agreed upon something.

Now, we'll just wait what happens when those québécoises start using Flyspray...

Psychokiller1888:
Well, I never completely checked the fr file that was there before.

supersquared:
Yep, I think the original translation was made by an auto-translator, and then checked by you, but not completely.

Sorry for the "poor translation" but it was true for some parts, and surely not created by a human.

I still working on it to have a "perfect french" translation ;)

PR #388, opened 2015-08-20
theme css update

supersquared:
- modified #taskdetails h2 & h2 input so it fill the width of the browser window

### DIFF
--- a/themes/CleanFS/theme.css
+++ b/themes/CleanFS/theme.css
@@ -562,7 +562,8 @@ form #taskdetails #taskdetailsfull {
   border-radius: 3px;
 }
 #taskdetails #navigation {float: right;padding:6px;}
-#taskdetails h2 {color: #555;display:inline-block;}
+#taskdetails h2 {color: #555;display:block;}
+#taskdetails h2 input {width:99%;}
 .tags span {
     background: #eee none repeat scroll 0 0;
     border: 1px solid #ddd;


peterdd:
I made it inline-block, so tags can float beside the heading in future if there is enough space in the browser window. will change html structure a bit in future.

What is the problem with inline-block here? (screenshot with annotation please?)
used for view task , edit task and new task.

https://github.com/supersquared/flyspray/commit/52a7d6454032aa76b6cca0e6c9926360ea0d6c9f

supersquared:
ok for tags.

The problem is the summary textbox, see the attached screenshot, it's a bit small I think

screen shot 2015-08-20 at 13 49 58

peterdd:
I see, how about min-width:300px; or width:300px; ?
I don't wanted the input field looking too big on big monitors, the user should enter short, but precise task headings.

supersquared:
I totally understand your point, but I think you should not fix the width. if you want to constrain the user input, maybe you could implement the maxlength for this field ?

peterdd:
The maxlength attribute is set to 100 in the input tag, so that constraint still exists and we shouldn't force a smaller size. Sometimes a longer summary is needed. So the visible input size is more type of a subtile recommendation. (i thought)

It could in future enhanced by javascript checking the input of user and show a hint if input longer than x chars. (like here on github.com when writing pull request heading >50 chars)

But I prefer jsless solutions if possible.

Hmmm....

peterdd:
ok, i made the input summery full size again, but dropped the h2 around.

PR #402, opened 2015-08-28
Use standard SQL, not MySQL dialect.

jahto:
And another small fix. Did I really forget to make a pull request yesterday?

### DIFF
--- a/includes/class.notify.php
+++ b/includes/class.notify.php
@@ -573,7 +573,7 @@ class Notifications {
           -------------------------------
          */
 
-        $body = tL('donotreply') . "\n\n";
+        $body = tL('donotreply', $lang) . "\n\n";
         $online = '';
 
         // {{{ New task opened
--- a/includes/class.project.php
+++ b/includes/class.project.php
@@ -292,14 +292,14 @@ class Project
         {
                 global $db;
                 if ($pm) {
-                        $result= $db->Query('SELECT tag `tag_name`, 1 `list_position`, 1 `show_in_list`, COUNT(*) `used_in_tasks`
+                        $result= $db->Query('SELECT tag AS tag_name, 1 AS list_position, 1 AS show_in_list, COUNT(*) AS used_in_tasks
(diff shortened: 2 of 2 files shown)


jahto:
@Psychokiller1888 now you will receive notifications mostly in french. I just have to add the integration to google translator for comments and task descriptions...

peterdd:
Really? I think automatic translation of comments and descriptions goes too far and also adds complexity too. (which project is allowed to call google translator and so on..)

We have so many other things to do for Flyspray first. ( onpage notification! ;-) )
If someone really wants a google translation he could use a browser plugin.

jahto:
I think Google has it nowadays somewhat right, here's some snippets from the discussion about pull request 387 (https://github.com/Flyspray/flyspray/pull/387) translated:

psycho: Nein, das ist nicht ok .... Vor es vor oder oben ist. Ein Highlight ist Höhepunkt
supersquared: Beide sind gültig Ich denke, habe ich es erneut, um "Beweise" vorlegen ich kurz
...
psycho: Einschaltstrom muss, ist ein Teil von denen immer htaccess es an der Wurzel der FS
supersquared: "Einschaltstrom" nicht auf Französisch vorhanden sind, gibt es "Trigger", aber es macht keinen Sinn, in diesem Zusammenhang zu machen

I hope you find these examples at least funny, although might not really understandable (now are you already laughing and rolling on the floor?). But maybe after 10 years... besides, as you maybe have noted, I modified TODO list for 1.0. There's still that asynchronous sending of notifications that should be done first. I saw some hints in current swiftmailers code that its capable of doing that. Just don't know yet how.

Onpage notifications then. We should get rid of most of the javascript cruft the project has acquired during the years and finally replace it with something else. I'm in favour of angularjs, because that's the only framework I have time to learn and will be using in another work project.

jahto:
Oh, had to try Google translator with our comment too. It actually made somewhat understandable translation to finnish. And it make you sound like Yoda in Star Wars.

PR #418, opened 2015-09-13
One last final change to the upgrader. Let the user decide.

jahto:


jahto:
Now I know that @Psychokiller1888 and @peterdd can do the translation to french and german, but if @jordanmendler could spare a few minutes of his time and translate what I wrote to comprehensible english.

PR #420, opened 2015-09-13
Fix undefined variable

landrok:


peterdd:
rejected because incorrect/no problem for current github master.

PR #427, opened 2015-09-21
Update tag class name

landrok:
Namespace for a tag is naturally .tag, then you can add a custom class with class="tag t_123".
.t_123 has a special custom role and should be only added to default .tag style when needed, overriding default style.
Some other reasons are clearness, reuse and less rendering costs than regex.
Proposition for a respective update for details.view.tpl will be pulled shortly.

### DIFF
--- a/themes/CleanFS/theme.css
+++ b/themes/CleanFS/theme.css
@@ -596,10 +596,8 @@ form #taskdetailsfull {
 }
 #taskdetails #navigation {float: right;padding:6px;}
 #taskdetails h2 {color: #555;display:inline-block;}
-/* t_ now reserved classes for tags with their tag_id like t_123
- Couldnt use t123 due limited regex in css and some other classes start with 't' 
-*/
-.tags span, [class^='t_'] {
+/* t_ now reserved classes for tags with their tag_id like t_123 */
+.tags span, .tag {
     background: #eee none repeat scroll 0 0;
     border: 1px solid #ddd;
     border-radius: 3px;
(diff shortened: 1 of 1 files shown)


peterdd:
I just wanted the htmlcode short.

But I haven't done a performance comparison between your prefered version (which I had first too in mind when adding tag styling) and the regex variant.

Beside from the regex usage I see no difference. Why is non regex version more reuseable than the regex version?

landrok:
Hello Peter, 
First, I know that this feature is in dev. It is just a proposition of patch, something like a try to give you some inspiration.
Then, ^='t_' is too general, not as indicative.  It is not easily reusable because it's reserved :) It has not enough semantic sense- think about new dev who are coming. More do you think it's a good way to use 2 consecutive global regexs on every page, just for a tag customization?

If you want to keep, the html and css short (working at Google ?), here it is 2 more reasonable versions:
```
/* with regex */
.tags span[class^='t_']
/* more supported and compliant seeing where the html is on the page */
.tags span{}
```

Even better:
```
/* position/sizing in details  */
#taskdetails .tags span{}
/* position/sizing in tasklist  */
#tasklist .tags span{}
/* theming for all context */
.tags span.t_1{}
.tags span.t_2{}
.tags span.t_3{}
```
HTML is preserved, no need for regex and no need for .tag too :)

Finally, I saw that that integration has started in the task list (/index.php?project=1), the changes I proposed was going in this direction. So, what's your opinion?

peterdd:
Let us keep this to pull requests open.

My head is full of nonflyspray related workload (small ecommerce company, xmas business coming) and I need to sleepover it.

landrok:
+restored

peterdd:
merge conflicts resolved manually

PR #428, opened 2015-09-21
Update tag class name

landrok:
Here is a little update related to the previous css pull.


peterdd:
I'm still a bit unsure how we want manage custom tag styling by the user (and other properties like priority, severity, category, tasktype)

I have added a dbfield to the list_tag table, thought with this I could reuse a tag styling for different tags like having 10 predefined coloring styles for the background and border (so it doesn't look too ugly) and some (font)icons
and user can combine them in the class field as "t_yellow fa-bug" in the list manager.

If the field is empty, the basic tag styling is used.
And if the dbfield class contains something like "auto", the t_123 style is added (or in your version "tag t_123") to the tag-tag.

Well just ideas, wanted at least the db change for tags into the Flyspray1.0, so we and user can play with it.

landrok:
+restored

PR #433, opened 2015-09-23
Fix reports URL in non rewrite mode

landrok:
A missing project id in CreateURL function causes 2 bugs:

**1 - Context:**
- URL rewriting desactivated
- a project id is specified (we are NOT on a global view) such as /index.php?project=4
- CreateURL has no project_id in "reports" case
- We want to see the "reports", so we are clicking on "Reports" top menu

Actual behavior: project id is lost during this action, reports for another project are shown.
Expected behavior: To keep project id of current project and show the reports for this project.

**2 - Context:**
- URL rewriting desactivated
- a project id is NOT specified (we are on a global view) such as /index.php
- CreateURL has no project_id in "reports" case
- We want to see the "reports", so we are clicking on "Reports" top menu

Actual behavior: project id 0 is lost during this action, reports for a specified project are shown.
Expected behavior: To keep project id 0 and show the reports for all projects to which I have rights.

This patch changes actual by expected behaviors for twice.

### DIFF
--- a/includes/class.tpl.php
+++ b/includes/class.tpl.php
@@ -960,7 +960,7 @@ function CreateURL($type, $arg1 = null, $arg2 = null, $arg3 = array())
             case 'myprofile':
             case 'register':
             case 'reports':
-            	$return = $url;
+            	$return = $url . '&project=' . $arg1;
             	break;
             case 'mytasks':
             	$return = $baseurl.'index.php?do=index&project='.$arg1.'&dev='.$arg2;


peterdd:
Please consider that the switch-case construct of php works a bit special and not so intuitive:
This is done because it works in C the same way and php took inspiration from that syntax too.


$return = $url;

is executed for all case 'xxx': above that have no break;

```php
case 'lostpw':
case 'myprofile':
case 'register':
case 'reports':
    $return = $url;
    break;
```

So you must always be careful within switch-case constructs and keep an eye on the break;

peterdd:
Well, until now I never thought about project specific event reports.

The title above the form on reports page should then at least be adjusted to "Reports for  xxx"
to make the context clear for the user.

And it brings also up this question:
  * Whats up with the tasks that are moved between projects?
  * Whats up with task relations between 2 projects?

related to FS#1950 on bugs.flyspray.org

landrok:
Hello, I don't understand why you are speaking about break instruction. I did not modify how break is used.
The only purpose of this patch is to specify project_id in url.
Impacts on other cases ['lostpw','myprofile','register'] urls are minimal: it puts a null project id or a 0 among context.
There is a few minutes, I pulled a second version which lets the other cases exactly as they were processed before. -> github #433

landrok:
Reports is one of the most important part of a project management tool.
Everyday we have these recurring questions: what was/is/will the workload of my teams? What efforts was done? How much it really costs? And obviously, how to improve/optimize that costs?

Combined with "effort" feature, I think it is/could be the decision helper part of the software but it is very far from this patch's purpose.

You're right, "Reports for xxx" would be more comfortable. For my everyday uses, I refer to the global title ("All projects" or "xxx") to quickly see on which project I am.

For the questions of task moves between projects and how fs should handle that in reports module, I think we could do as dependencies were seen in 0.9.9.7. It is just an idea since I did not make a complete check of this feature. Thanks for questions but it's too far from this patch too !

PR #440, opened 2015-09-28
add project_id in query params

landrok:
Related to #439
- add project_id in query params for callbacks/searchtask.php


peterdd:
Why not document.getElementsByName('project_id')[0].value ?

There was some short discussion to think of review of all javascript parts in Flyspray, so it could be possible prototype is not available in future.

PR #441, opened 2015-09-28
Fix a PHP Notice + PDO call

landrok:
Related to #439 and Peter's comments (Thx)
- add removed trim
- add loading of user profile
- add "can_open_task" access checking
- add project_id in where clause

Other problems:
- dokuwki not supported+js enabled => controls should be done again in saving flow because can not be done in this one (js event only)
- details is optional: the 2 fields (summary+details, now project_id) are always sent. Optional points filled status and has no impact in control. If a future version plans to remove fields at submission when not required, we should patch it when done ?

### DIFF
--- a/js/callbacks/searchtask.php
+++ b/js/callbacks/searchtask.php
@@ -35,7 +35,7 @@ $sql = $db->Query('SELECT count(*)
 		   FROM {tasks} t
 		   WHERE t.project_id = ? 
 		   	AND t.item_summary like ? 
-		   	AND t.detailed_desc like ?'
+		   	AND t.detailed_desc like ?',
 		   $params);
 $sametask = $db->fetchOne($sql);
 echo $sametask;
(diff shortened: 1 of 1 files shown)


peterdd:
Why do you removed the tags css pull requests? I said I want to just keep it open until a decision. (read:  have time to work on that as a whole..)

landrok:
I did not voluntarily... I thought it was you!
I made a "big clean delete patch-x branches party" and CSS pulls were among them.
-> Restored a few minutes ago.
I haven't work on tags since it's a feature indev, I don't want to disrupt first implementation (best effort).

peterdd:
lol, I thought that. ^^
Ok, now we know that pull request from branch-x to master is removed when branch-x deleted. :-)

peterdd:
One question: The check uses $proj , but the project_id from the post is used later in the query.

What if $proj id differs from the post project_id?
Could it possible the check is done on $proj (id=2) and then in the query a project_id=3 is used?
So the can_open_task() is done on the wrong project id?

PR #445, opened 2015-10-10
Flyspray max file size for attachments

kandran:
When memory_limit or post_max_size or upload_max_filesize is set to unlimited value size doesn't be used for the computation of the attachments max size -  so if the three variable are set to -1 (unlimited) there are no attachment allowed but config allow them so in these case we could set value to a big number (php_int_max).


peterdd:
Thx.

Issue #56, opened 2014-10-02
next release after 0.9.9.7 ?


Sajaki:
hello, 
I see alot of merged PR's  but can you tell when the next release is done and what version that will be ?

msoky:
when it will be?

jordanmendler:
No specific plans as of now. Code contributions are welcome

PR #451, opened 2015-10-23
add project_name as request parameter

mulle-nat:
Hi

With this patch you can specify ?project_name=<project_title> instead of the ID with ?project in the URL.

I also think
line #56 in index.php is wrong

``` php
   list($proj_id, $orig_name, $file_name, $file_type) = $task;
```
and should rather read 

```php
   list($project_id, $orig_name, $file_name, $file_type) = $task;
```

but I didn't test that.

P.S. Caveat: I don't know PHP :)

### DIFF
--- a/header.php
+++ b/header.php
@@ -87,11 +87,21 @@ if (in_array(Req::val('do'), array('details', 'depends', 'editcomment'))) {
 
 if (!isset($project_id)) {
     // Determine which project we want to see
-    if (($project_id = Cookie::val('flyspray_project')) == '') {
-        $project_id = $fs->prefs['default_project'];
+    if ( Req::val('project_name') != '') {
+         $result = $db->Query('SELECT project_id
+                               FROM {projects} WHERE project_title = ?', array( Req::val('project_name')));
+         $project_id = $db->FetchOne($result);
     }
-    $project_id = Req::val('project', Req::val('project_id', $project_id));
+
+    if (!isset($project_id)) {
+        if (($project_id = Cookie::val('flyspray_project')) == '') {
+            $project_id = $fs->prefs['default_project'];
+        }
+        $project_id = Req::val('project', Req::val('project_id', $project_id));
+    }
 }
+
+
 
 $proj = new Project($project_id);
 # no more project cookie!
--- a/index.php
+++ b/index.php
@@ -104,7 +104,7 @@ if ($conf['general']['output_buffering'] == 'gzip' && extension_loaded('zlib'))
 $page = new FSTpl();
 
 // make sure people are not attempting to manually fiddle with projects they are not allowed to play with
-if (Req::has('project') && Req::val('project') != 0 && !$user->can_view_project(Req::val('project'))) {
+if ( ! $user->can_view_project(Req::val($project_id))) {
     Flyspray::show_error( L('nopermission') );
     exit;
 }
@@ -164,7 +164,7 @@ $sql = $db->Query('
 	ORDER BY project_is_active DESC, project_title'
 );
 
-# new: project_id as index for easier access, needs testing and maybe simplification 
+# new: project_id as index for easier access, needs testing and maybe simplification
 # similiar situation also includes/class.flyspray.php function listProjects()
 $sres=$db->FetchAllArray($sql);
 foreach($sres as $p){
@@ -193,7 +193,7 @@ $page->pushTpl('header.tpl');
 if (!defined('NO_DO')) {
     require_once(BASEDIR . "/scripts/$do.php");
 } else{
-    # not nicest solution, NO_DO currently only used on register actions 
+    # not nicest solution, NO_DO currently only used on register actions
     $page->pushTpl('register.ok.tpl');
 }
 $page->pushTpl('footer.tpl');


peterdd:
1. Why do you want that? A project name can always be changed. If users bookmarked it the link fails in future.
If you really want nicer urls use url_rewrite feature and use .htaccess


2. Think in line 56 is not a real issue. The var is never used there, just a placeholder to set the other 3 vars there correct.

codeon-nat:
If you don't like it, don't use it :) 

I have a setup, that sets up git repositories, gitlist homepages, gitstats and now also the bugtracker. A bit like a homegrown github.

Linking by ID isn't possible, because I have no place to store the reponame <-> projectid association.

peterdd:
In this case I suggest writing a feature request on bugs.flyspray.org

Describe as rich as possible what you want and what is required.

Flyspray currently hasn't a plugin/hook system yet. But this seems a good candit if we have hopefully in future.

The cookie for flyspray_project is obsolete and will be removed in future.

peterdd:
And is API related too, which also doesn't exist yet for Flyspray.

mulle-nat:
??? It's a pull request, the code is already there and working for me.
Or do you mean the automatic project creation ? I have code for that too :)

peterdd:
I just want avoid workarounds or any more shortcomings. And this looks like a workaround for a problem of your own software.

If you have some API-like stuff for Flyspray, why not share this? Is it REST or SOAP-based?

mulle-nat:
[gitlist](http://gitlist.org/) is not my own software.

I doubt you want this as it's not pretty, but here it is, in case someone else has a similiar problem.

```php
 <?php

$config = parse_ini_file( 'flyspray.conf.php', true);
$secret = $_REQUEST['secret'];

if( strlen( $config[ "oauth"]["mulle_secret"]) == 0)
{
   die( 'Need a config variable "mulle_secret" under "oauth", in flyspray.conf.php');
}
if( $secret != $config[ "oauth"]["mulle_secret"] )
{
   die( "I don't trust you." . $secret);
}


$conn = mysqli_connect( $config[ "database"][ "dbhost"], $config[ "database"][ "dbuser"], $config[ "database"][ "dbpass"], $config[ "database"][ "dbname"]);
if (mysqli_connect_errno()) 
{
    die("Connection failed: " . mysqli_connect_error());
}

$project_title     = mysqli_real_escape_string( $conn, $_REQUEST['project_title']);
if( strlen($project_title) == 0)
{
   die( 'Need at least a project name');
}

$intro_message     = mysqli_real_escape_string( $conn, $_REQUEST['intro_message']);
$notify_email      = mysqli_real_escape_string( $conn, $_REQUEST['notify_email']);
$feed_img_url      = mysqli_real_escape_string( $conn, $_REQUEST['feed_img_url']);
$feed_description  = mysqli_real_escape_string( $conn, $_REQUEST['feed_description']);

$sql = "SELECT project_id FROM flyspray_projects WHERE project_title = '" . $project_title . "';";

$result=mysqli_query( $conn, $sql);
if( ! $result)
{
    $error="Error: " . $sql . "<br>" . mysqli_error( $conn);
    mysqli_close($conn);
    die( $error);
}


$row = mysqli_fetch_array($result);
$value = $row['project_id'];

# does not work
# $count=mysqli_num_rows( $conn);

if( $value != '')
{
    mysqli_close($conn);
    die( "project " . $project_title . " already exists");
}


$sql = "INSERT INTO flyspray_projects 
( 
   project_title, 
   theme_style,
   intro_message, 
   project_is_active, 

   visible_columns, 
   visible_fields,
   others_view, 
   others_viewroadmap, 

   notify_email, 
   notify_types, 
   feed_img_url,
   feed_description, 

   notify_subject, 
   lang_code
)
VALUES 
(
  '$project_title',
  'CleanFS',
  '$intro_message',
  1,

  'id category tasktype priority severity summary status progress',
  'tasktype category severity priority status private assignedto reportedin dueversion duedate progress os votes',
  1,
  1,

  '$notify_email',
  '1 2 3 4 5 6 7 8 9 10 12 13 14 15 16 17',
  '$feed_img_url',
  '$feed_description',

  '%p: %s %a (%u)',
  'en'
);";


if( ! mysqli_query( $conn, $sql))
{
    echo "Error: " . $sql . "<br>" . mysqli_error( $conn);
}
else
{
   $key = mysqli_insert_id($conn);
   echo $key;
}


mysqli_close($conn);

?>
```

peterdd:
Well, that quick project creation doesn't use/respect flyspray permission system.
You could use  'admin.newproject' section of includes/modify.inc.php of flyspray.



I took a look at git and gitlist today.

You could also store the mapping git->bugtracking for instance with 
```
; link a repository to a flyspray project id
[bugtracking]
mynewproject=/flyspray/proj1
; or on external server:
mynextproject=https://example.com/proj1
```
in gitlists config.ini.

In  Application.php constructor add
```
$this['bugtracking'] = $config->getSection('bugtracking');
```

And show the link for example in navigation.twig: 
```
<a href="{{ app.bugtracking[repo] }}">Tasks</a>
```

Or store it in git config itself? (https://git-scm.com/docs/git-config)

mulle-nat:
Yes that would be a nicer way of getting the information into gitlist and lets me migrate easier to something else in case it's needed.

peterdd:
Are you still registered on bugs.flyspray.org?

Interested in planning API for Flyspray? I would prefer the use of some standard methods (REST,SOAP, do not reinvent the wheel) + using Flyspray's user and permission system that restricts/controls the power of the APIusers + usage limits (to avoid spamming, excessive resource usage, crazy apiusing bots)

So for instance use an existing user or create a 'api-user' and setup apisecrets (if not using user/password) and set permissions for this user (e.g. put him into a global-norights-group and a project group that only can do comments like a automated information from github when a commit refers to a FS# task.).

mulle-nat:
I would like to, but I have really no spare time at all. 

BTW, what you detailed would be a nicer way of managing the ?project_name= idea within git and gitlist. But it wouldn't solve my original problem. The  ?project_name= and my script solves it, to my current satisfaction.

Also quite frankly the URL looks nicer this way, compared to flytrap/index.php?project=6. Yes there is URL rewriting, but that's even more work I would have to do.

PR #454, opened 2015-10-27
Norwegian Language File updated no.php

jack-reiersen:
Norwegian Language File updated 
no.php

Will use it from now on and check for typos.


peterdd:
Looks a bit weired. for instance line 71 'anonview'. Wrong editor settings?
Please redo with correct utf8.

jack-reiersen:
Yes. I see now. copy paste to fast. will update and comeback later.

PR #457, opened 2015-11-01
Fix for FS#2075, users with empty jabber ID don't get notifications n…

danoh:
Fix for FS#2075, users with empty jabber ID don't get notifications now, so this fixes security problem,
but there is still problem with the notification logic, people with jabber ID and email are getting email
notifications parallel/instead of jabber notifications.


peterdd:
Merged my older fix now after a little testing.

People with email and jabber should get notification on both channels if they selected NOTIFY_BOTH in their account.

PR #463, opened 2015-11-12
Update tag class name

landrok:
- tag_id is not appropriated because it forces having one class per tag
- t prefix or other one (cus-tag-*) should be defined when adding/updating a new tag style class
- by this way, if no dedicated style has been defined, there no "t" class, no space in class attribute - keeping html short ;)
- this approach seems to be consistent with flyspray_list_tag.class nullable field

@peter In order to discuss about the tool to use for styling tags, we should speak on fs#2012 ?

### DIFF
--- a/themes/CleanFS/templates/details.view.tpl
+++ b/themes/CleanFS/templates/details.view.tpl
@@ -683,7 +683,7 @@ function quick_edit(elem, id)
 	<h2 class="summary severity<?php echo Filters::noXSS($task_details['task_severity']); ?>">
 	FS#<?php echo Filters::noXSS($task_details['task_id']); ?> - <?php echo Filters::noXSS($task_details['item_summary']); ?>
 	</h2>
-	<span class="tags"><?php foreach($tags as $tag): ?><span class="tag t<?php echo $tag['tag_id']; ?>"><?php echo Filters::noXSS($tag['tag']); ?></span><?php endforeach; ?></span>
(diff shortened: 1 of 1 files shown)


peterdd:
Hi landrok, I'm writing displaying tag styles on pm and admin common.list.tpl at the moment.

- Yea, one class per database tag_id. If not set any css for a t1234, browser don't care, right?
I think the list of tags shouldn't be too high to keep it simple for users. Adding/Removing tags to tasks should/could support correct usage by autocompletion/dropdowns, so people do not type different tag names for same. (only if adding new tag names allowed)
- It is now 2+iddigits per tag, I think we can ignore this compoared to the rest of html output?
- The rules could be grouped together in css, like 
.t4, .t5, .t6{ background-color:red;}
.t7, .t8, .t9{ background-color:blue;}

- The usage of class-field in database could be a second easy way for endusers to configure the style by selecting predefined tag classes. The result could then be:
```
<i class="tag t22 tlegs tgreen">walk</i>
<i class="tag t22 tbicycle tgreen">cycle</i>
<i class="tag t23 tcar tred">drive</i>
<i class="tag t24 tairplane tred">fly</i>
```

landrok:
Ok, sorry for disturbing implementation.
I will have a look at that feature later when implementation is more advanced.
I let you close that pull if no other comment.

peterdd:
No problem, good to see if other people keep an eye on what I do. :-). I probably add the 'class'-field also as you proposed.

PR #468, opened 2015-11-14
Add baseurl to global js scope

landrok:
See comment on FS#1737-IC

PR #472, opened 2015-11-14
Update PHP $baseurl to JS baseurl

landrok:
#468
FS#1737-IC

### DIFF
--- a/themes/CleanFS/templates/admin.prefs.tpl
+++ b/themes/CleanFS/templates/admin.prefs.tpl
@@ -190,7 +190,7 @@ function ShowHidePassword(id) {
           <label class="labeltextarea" for="intromesg"><?php echo Filters::noXSS(L('mainmessage')); ?></label>
           <?php if (defined('FLYSPRAY_HAS_PREVIEW')): ?>
             <div class="hide preview" id="preview"></div>
-            <button tabindex="9" type="button" onclick="showPreview('intromesg', '<?php echo Filters::noJsXSS($baseurl); ?>', 'preview')"><?php echo Filters::noXSS(L('preview')); ?></button>
+            <button tabindex="9" type="button" onclick="showPreview('intromesg', baseurl, 'preview')"><?php echo Filters::noXSS(L('preview')); ?></button>
           <?php endif; ?>
           <?php echo TextFormatter::textarea('intro_message', 8, 70, array('accesskey' => 'r', 'tabindex' => 8, 'id' => 'intromesg'), Post::val('intro_message', $fs->prefs['intro_message'])); ?>
         </li>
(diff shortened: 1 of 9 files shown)


peterdd:
OK, give me some time to review if replacing a stringconstant (viewed from js-side) by a js variable baseurl do not harm. (like another script overwrites the var baseurl)

But having a var baseurl available could also help fixing bug with the lightbox, which doesn't load some ui-images if urlrewriting(niceurls) is activated in flyspray. (js/lightbox/js/lightbox.js line 49) 
```
 fileLoadingImage:        'js/lightbox/images/loading.gif',     
 fileBottomNavCloseImage: 'js/lightbox/images/closelabel.gif',
```

landrok:
For naming conflicts, we could eventually use a prefix like `fs_baseurl` instead of `baseurl`.
Another solution, in a second pass, could be using HTML "base" element in head part of the page.
```
<base href="'.$baseurl.'" />
```
Why in a second pass ? 
I think that the current topic is only to extract inline js and to avoid php/js melting.

Problems with url rewriting should be resolved with htaccess updates?

But take your time, we're free/libre here :)

peterdd:
'Another solution, in a second pass, could be using HTML "base" element in head part of the page.'
Yea, had same idea before: https://github.com/Flyspray/flyspray/tree/basehref-for-urlrewriting

But still a bit unsure if that's best solution and if it breaks anything.

landrok:
Yes, this is why it may be done in a second pass.
Base element is mainly useful for relative url, eventually breaking some existing workarounds. This is why I specified that this topic was to extract inline JS (Focused on FS1737).

I think this is one of the priority, before writing new features. Extracting a maximum of JS inline code should permit us to write new features with JS&CSS frameworks, standardizing code and avoiding writing same features several times. Maybe I am wrong but the work is done on https://github.com/landrok/flyspray/tree/PHP/JS-meltings-clean-scriptinlinemarkup branch.

There are still 5 parts to rewrite: 
- common.list.tpl: used in tag feature, so I let you write before reforge new way. quick rewrite, same way that common.cat.tpl
- depends.tpl: waiting total rewrite of the template according to the head comment
- common.userselect.tpl: needs some more analysis, waiting if I am wrong on other patches according to official fs team
- details.edit.tpl: inline js is a little hack, 2 * one line, rewrite will be quick
- loginbox.tpl: inline js is a little hack, rewrite will be quick

peterdd:
At least the Restler project uses a baseUrl in its code. 
Steven tried some work to add it in 2014: https://github.com/Flyspray/flyspray/tree/Rest_Web_Services
But I haven't tried and absolutely unsure if that is the right tool for a future Flyspray API.

Maybe the base makes everything simplier for us. Will test.

landrok:
```
<base href="'.$baseurl.'" />
```
And some inline JS rewrites deprecate this JS approach. FS1737 may be done later.

PR #476, opened 2015-11-14
Fix a PHP Parse error

landrok:
PHP Parse error:  syntax error, unexpected '}' in flyspray/themes/CleanFS/templates/roadmap.tpl on line 57
This error was first avoided by `// }` at the end of the foreach and then hidden by a `<?php } ?>` later in the code. See 2 parts of code which have been modified for details.

### DIFF
--- a/themes/CleanFS/templates/roadmap.tpl
+++ b/themes/CleanFS/templates/roadmap.tpl
@@ -66,7 +66,7 @@ allTasks<?php echo Filters::noXSS($milestone['id']); ?> = [<?php foreach($milest
             }
             $effort = null;
         }
-    // }
+    }
     ?>
     <br />
     <?php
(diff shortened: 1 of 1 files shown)


peterdd:
Why the current code has a parse error here:  "// }" ???
It seems jahto included the bottom parts into the ' if ($proj->prefs['use_effort_tracking']) {'

So if the admin or project manager completely disables 'effort_tracking' for a project, 
the 'totalestimatedeffort' and 'currenteffortdone' isn't shown, regardless of user permissions.

landrok:
Because `// }` is dead code, so if you uncomment this, there is a parse error.
There is one of the two close brackets which is not needed. first or second one.

peterdd:
I think jahto let the closing bracket as a comment in here to point out his change now includes the two perm checked echo's . Maybe he wasn't 100% sure if projects 'use_effort_tracking' overrules view_estimated_effort and and view_current_effort_done permission for every user.

peterdd:
I have to close it, because it changes logic. And toplevel.tpl and roadmap.text.tpl use the same logic.
You can remove the commented out bracket, but not the last bracket.

PR #485, opened 2015-11-26
Polish lang update

Krzysiu:
1) Flyspray'a > Flyspraya, Flysprayu
In Polish we use apostrophe only if the last vowel is mute. In this case "y" is not mute.
2) Byłeś(-aś) > non gender forms
Some words in Polish are gender depended. Things like "Byłes(-aś)" means "byłeś [for men] byłaś [for women]. Polish linguist, well known expert, suggests that the best way is to use non gender forms, like "Was password forgotten?" instead "Did you [gender depended word] forgot your password?". In case when it's not possible, he suggests using form "byłeś / byłaś" (most natural, easily readable), but I didn't use it, as there's problem with length of such strings, especially with longer words.

I don't know which is better - "byłeś(-aś)" or "byłeś(aś)", yet most of translation has "(-aś)", so I unified it, where that other form was used. Congratulations for translators for using two forms instead one - for men! That's very noble, modern and pro equality :) I see such forms rarely, so thanks!
Source: http://sjp.pwn.pl/poradnia/haslo/Czy-bylaby-pani-uprzejmy-a;10969.html
3) Simplification of longer strings
Some strings were over-complicated or there were non-typical grammatical syntax, so for accessibility I made it simpler with exactly same meaning. 
4) Minor grammar fixes and typos
5) Effort
Effort is translated as "job", I changed it to "time of job", as it's closer to that English word in this case.


peterdd:
Thank you for reviewing polish translation.
I think for discussion polish grammar contact the guys who contributed to polish translation (xmesaj2, ..?) see https://github.com/Flyspray/flyspray/commits/master/lang/pl.php

The amount of polish translation strings is currently 87%, so there are 124 strings yet that are untranslated (fallback to english).

Flyspray comes with a builtin translation help tool.
You can login into a fresh (github master) flyspray install as admin to view and the translation area: 
/index.php?do=admin&area=translations   (overview)
/index.php?do=langedit&lang=pl  (button show missing)
/index.php?do=langdiff&lang=pl  (missing and quirks like deprecated identifiers)

After finishing (confirm all changes button) you can copy & paste the /lang/pl.php here on github.

PR #490, opened 2015-12-16
Change tabs highlighting

landrok:
When a search is done (tasklist context) and $_GET['dev'] is not empty, the tab "mytask" is highlighted whatever be the dev value.
When a search is done (tasklist context) and $_GET['dev'] is empty, no tab is highlighted.

With that patch, 
the tab "mytask" is highlighted only when $_GET['dev'] == $user->id. (not username)
Otherwise, for tasklist context, this is "tasklist" tab which is highlighted.


peterdd:
Thx.

PR #493, opened 2015-12-29
User-Notification-Limitation

rkCSD:
For our company, we needed a limitation of user-notifications. Our customers can't login to flyspray, because they want to get status-emails.

Our goal: Adding customers as "Notification-Users". The customers should only get mails, if the task "itself" changes (changing the progress, edit the task-description,... (notification-type 1-4)). Emails for new comments or other things (notification-type >4) should not send to the customers (customers should not know every comment ;-) ). A customer have a "silent-account" at flyspray: It's a normal user which is in a group without any rights.

We added a "User-Notification-Limitation"-Checkbox at the notification-tab.
screenshot

If you checked "Limited", the notification-user only get emails for notifications, there are specified as type 1-4.

**Important:** A database-change was needed to add this functionality. We added the new field in the Installation-XML *ONLY* for version 1.0-rc. It's not contained in upgrade-scripts in lower versions. We weren't sure, what was the best way to doing this.


WooterTheTroubleshooter:
Personally I would make this a multiselect dropdown where you select any of
the notification types you want this user to receive, another company might
only want to sent their users when a task has been completed.

peterdd:
- When adding a new dbfield, it must also be done in the upgrade.xml, not only in the flyspray-install.xml. So people do not get in trouble if  they upgrade from older version or update a 1.0-xxx.

- This patch can only be seen as temporary solution. The notification system needs a review (2016)
probably starting with related FS#1812, probably a separation of user action php process and sending of notifications over different channels (email, jabber/xmpp, .. and other future channels)
review of notification types see @WooterTheTroubleshooter comment, configurable grouping of notification sending so users do not get too much spammed by small task updates ("quickedit"-feature - not invented by me..) , overlapping between "history" and notifications.

Because Flyspray permission stuff is quite big now I really want some demo configurations documented with screenshots in future on www.flyspray.org.
Your needs are quite common (a company with internal staff and customers with limited access for own support-tasks).

- I'm currently stuck a bit with travis-ci/phpunit giving false-positive results, which is blocker for postgres_concat-patch as I haven't a working postgresql-setup at the moment and no active testers ( @jahto  where are you? ), which is blocker for some stuff I would like to finish for FS1.0. So be a bit patient with a merge to master (We currently avoid any db-change for FS1.0 if not absolute necessary)

rkCSD:
Maybe it's better to close this pull-request. I had make a mistake during the last commits. The latest commits shouldn't append to this pull request, sorry.

PR #495, opened 2016-01-16
Fix loss of project context

landrok:
- Fix loss of project context while attempting to edit a group (clicking on heading buttons to edit a group)
- Fix loss of project context while editing a group (selecting another group to edit or at form submission)


peterdd:
Thx for finding the bug.

Because a group is bound to a project, the project id can be read from db-entry and a project_id in the editlinks is redundant.

So we can detect the project_id in header.php I think..

landrok:
#499 did it. :+1:

PR #503, opened 2016-01-17
Add email format control

landrok:
Trying to add an user with a bad email (such as "test") throws the following swift exception: 
```
Completely unexpected exception: Address in mailbox given [test] does not comply with RFC 2822, 3.6.2.
This should never happend, please inform Flyspray Developers
```


landrok:
- Message is already defined
- A new PR will use Flyspray::check_email() method

PR #505, opened 2016-01-17
German translation: Fixes and additions

EifX:
Added some Translations to german language. I don't no why, but %d-Replacement doesn't work sometimes (I had this issue only in Task-Overview). I replaced this in german language with an alternative translation.


peterdd:
Yea, i started this with %d-replacement for better translations, but hadn't time/other priorities to finish it.

PR #507, opened 2016-01-19
Fix perms potential weakness

landrok:
Calls with a project identifier in array format are crashing the script with PHP warnings. Several do actions are concerned.
```
GET /index.php?do=reports&project[]=1
```
=>
```
Warning: Illegal offset type in /includes/class.user.php on line 86
```

### DIFF
--- a/includes/class.user.php
+++ b/includes/class.user.php
@@ -78,7 +78,7 @@ class User
     }
 
     public function perms($name, $project = null) {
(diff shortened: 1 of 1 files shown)


peterdd:
Mmmh, shouldn't this be handled by header.php around line 102?

landrok:
Indeed, it would be better to do so. #510

PR #509, opened 2016-01-19
Fixed design-bug in extended search

EifX:
- Percent Select-Field was too large
- Typo in german language fixed


peterdd:
Mmh, i made this with intention, so user sees every value of this dropdown.
But I see it doesn't fit with the other selects..

EifX:
Yes, I can understand you ;) But I think, you have two choices:
* Larger Search-Area: Every Search-Field must have a size of 12
* Smaller Percent-Area: Percent-Area with size of 8

I think, that's not a problem, that users must scroll the values up and down. They have to do it in the other fields, too. ;)

peterdd:
Third choice: 
```css
.search_select #percent > option {
    font-size: 8px;
}
```
but it is a bit too small and not perfect aligned. So maybe adding 
```css
.search_select select{
    height:150px;
}
```

too..

peterdd:
made now some changes to github master with the third choice. seems ok in firefox,chrome,ie,opera 
(desktop versions)
Please check if my changes are ok so I can close this pull request.

EifX:
I don't think, resizing the text is a "pretty" option. I think it's not harmonic to the other boxes.

Scrolling in this box should not be the problem. If you have more fields in "status" or "category", you have to scroll there, too. Why not in the percent-box?

peterdd:
Well, I hate the need to scroll option lists to see all options, especially when there are only 5-12 options. (I know mobiles use often their own native select ui as compromise for the small display, but still think it is not optimal.)

The difference to the other select boxes is that the percent select box has a fixed amount of options.
I had the idea of also style the option background with the percent$i classes - one of the reasons I use the new tpl_select() instead of tpl_options() template function here to add a css class to options.

But I see your point (harmonic). Maybe I move the font-size-css for that box to custom_example.css ..

PR #529, opened 2016-02-28
Add multiple email check

landrok:
Related to #504 Peter's comment and FS#1812

### DIFF
--- a/includes/class.flyspray.php
+++ b/includes/class.flyspray.php
@@ -1115,14 +1115,30 @@ class Flyspray
 
     /**
      * check_email
-     *  checks if an email is valid
-     * @param string $email
+     *  checks if one or several emails are valid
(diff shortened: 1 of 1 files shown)


peterdd:
That multiple email addresses thing should be postponed to the time when FS#1812 
database redesign is done.

Maybe doing the split by ';' inside a function check_mail() is not the right thing. What if you have an array of emails later you want check? Then you have to implode it as argument and then explode inside the function. Lets check_mail() only what it should do, but not more.

';' can be also appear in a quoted localpart of an email address, so the splitting just by ';' is a bit fishy. 

Unless there is are  real urgent requirements for the multiple email address feature I would like to drop it for FS1.0 release and just check the string as it would be a single address to check.

landrok:
I agree with postponing this check to a more mature implementation of multiple emails feature.

PR #534, opened 2016-03-02
fix local language

yetist:
Please see [FS#2106](https://bugs.flyspray.org/index.php?do=details&task_id=2106)


peterdd:
Not applicable this way, other language files are named lowercase_UPPERCASE.php like fr_CH.php

See  https://bugs.flyspray.org/2106

PR #535, opened 2016-03-02
locale CN uppercase

peterdd:
Hey chinese guys,
 
please check if my assumption is correct and give feedback.


yetist:
Yes, you are right, in Chinese language, should call setlocale(LC_ALL, "zh_CN.utf8").

PR #536, opened 2016-03-02
polish locale

peterdd:
Hey polish guys, 

please check if my assumption is correct and give feedback.


yetist:
I am Chinese, in my linux system, to run `$ locale -a|grep pl`, it will show:
>pl_PL
>pl_PL.iso88592
>pl_PL.utf8

PR #537, opened 2016-03-03
support i18n when install

yetist:
Please see [FS#2108](https://bugs.flyspray.org/2108)


peterdd:
Hi yetist, just use the current master as starting point for your cas work. It makes a possible merge later easier.

yetist:
This merge request is based on the current master now.
The cas work is not ready now, if it is ready, I will request to merge that.

PR #538, opened 2016-03-04
Install i18n

yetist:
Please branch used for i18n the install page. See  https://bugs.flyspray.org/2108


peterdd:
First, great that you continue work I started a year ago now (whoa, 1 year- time flights) I had no time for in the last months.

I prefer pull requests, that do changes only one thing.
It makes reviewing much easier and errors can be better tracked down to the point where a change introduced a bug/problem.

The task is https://bugs.flyspray.org/1849

peterdd:
* libchecktext without p-tags please. Avoid including wrapping tags in translation strings.
* in libchecktext is a typo: 
```
<string>  --> <strong>
```

yetist:
Use different flyspray-install-$lang.xml files just for locale language of  the initialize data.
On the maintenance issue, I suggest that the official release version does not provide any flyspray-install-$lang.xml files. This problem can be discussed later.

landrok:
Please, could you remove 85b5d50 "update fr.php file for translation" patch ?

- "en" values are fallback values, so it is not useful to have them repeated here.
- It makes harder to detect the need of translations.

I think b3821d3 "update de.php file for translation" patch should be removed too, for same reasons.

yetist:
"en" as fallback does not work when install flyspray, so I add the english values into "fr", "de" and "zh_cn", if it works, these values should be removed.
Please see line 7 of [i18n.inc.php](https://github.com/Flyspray/flyspray/blob/master/includes/i18n.inc.php#L7), here just load "lang/en.php", but when install flyspray, it should load "setup/lang/en.php" file as fallback.

landrok:
You're right. Maybe we could delete /setup/lang/* and merge setup translations with other application translations in /lang/. It could bring some advantages such as:
- better translation maintenance
- avoid reforging the load_translation methods (the hook to load from /lang is hardcoded)
- reuse some still existing entries & use dedicated to setup ones from admin toolbox

Did you try your updates this way ?

yetist:
OK, in i18n.inc.php, it use BASEDIR as base directory of lang/*, but visit http://domain/ and http://domain/setup/, the value of BASEDIR are different, is there any constant or variable defined to  indicator the absolute path of the top directory of flyspray program?

yetist:
Maybe define APPLICATION_PATH(value is BASEDIR) in includes/constants.inc.php is an easy way, then replace BASEDIR with APPLICATION_PATH in i18n.inc.php, but I'm not sure it is enough.

peterdd:
The idea of having a second setup/lang/ directory was, that the strings for the installer doesn't have to be loaded on every normal flyspray request. They are needed only once at install time (or upgrade) and then not again.

I agree that it does not make sense to copy untranslated en-values to fr.php , de.php .. 
It should just use the en-fallback if the translation do not exists in fr.php, de.php ..

Also I have to say that the installer as a whole can be reviewed, the step flow and texts can change, parts still use the old blue theme, that "do composerstuff with webinstaller"-part is highly experimental. So probability the default setup/lang/en -values changes are much higher then the en-values of normal lang/ directory.

landrok:
2 files are to be modified to solve BASEDIR conflict:
-  /setup/index.php
Replace BASEDIR by SETUP_BASEDIR in following 4 lines
l14
l15
l17
l994
Then add a ```define('BASEDIR', APPLICATION_PATH);``` after ```define('APPLICATION_PATH', dirname(SETUP_BASEDIR));```

- /includes/class.tpl.php
near from l99 replace BASEDIR by SETUP_BASEDIR
```
// This is needed to catch times when there is no theme (for example setup pages)
            require BASEDIR . "/templates/" . $_tpl;
```

when no theme is defined, it redirects to setup templates

After that, the cross context hook would be complete.

landrok:
@peterdd I understand this idea. The cost of the overhead (of loading lang strings) in every request is not important because of PHP opcache. The staying cost could be RAM, it is also not very important even if the strings dedicated to setup process are twice more as now.
Moreover, we can reuse strings in twice contexts, which could be so useful.

To avoid multiple changes during setup recoding, we could create a comment part (as @yetist did)

/****** setup process start ******/
[...]
/****** setup process end ******/

peterdd:
@yetist 

1. There is a (unwritten?) convention, that the translation key strings do not use ' ', '_' or '-', only ascii [a-z]. Advantage: coders have not to think about if its written with a 'separator' or not.

2. Is it required to translate configuration variables like safe_mode?  The value is for people who understand php. A localizatio of the config value would probably more puzzle all peoples, who who undestand php config and who who not.

3. Could you please split this pull request at least the following parts. There is feature freeze and do not want risk any problems due new features.

a. only changes that replaces hardcoded english with translation strings and adding them to setup/lang/en.php (and maybe add your zh_cn.php translations)

b. logic changes in core files

yetist:
@peterdd Code has been updated, please review.

peterdd:
part one is now merged, so I close this.

PR #541, opened 2016-03-05
Fix a notice

landrok:
This notice occurred when class name is set to NULL for a tag (list_tag table).

```GET https://bugs.flyspray.org/index.php?project=0&do=index```


peterdd:
related is FS#2102

and the sql querybuilding part related to tags in includes/class.backend.php function get_task_list()

PR #542, opened 2016-03-06
Add label for attributes

landrok:


peterdd:
I have no problem to merge this, but what is the gain here? Do some mobile Browsers popup the select when clicking on the label?
In Firefox on Desktop the select only flickers/highlights/set 'active' a bit and nothing more happens.

PR #544, opened 2016-03-08
Installer i18n part one

yetist:
- Replace hardcoded english with translation strings in template files

- Update English language file.

- Add Chinese translation file


peterdd:
@all: One more thing for translations:

I prefer translating verbs and adjective strings like 'enable' also as lowercase like 'enable'=>'enable'.
If you want capitalize it in a template, there is the css style "text-transform:capitalize;" that can be used the css file classes.

peterdd:
On the other side: 
In english adjectives, verbs, nouns are lowercase. And capitalized on headings.

But in german are nouns capitalized everytimes and the difference between verbs/adj. vs noun is important for reading (in my opinion).
And the other words are capitalized only at start of a sentence. 
(Although capitalizing verbs/adjectives on table columns cells is done also on german websites - which I find terrible because of the important meaning of lowercase/uppercase words)

PR #546, opened 2016-03-10
Cas authentication for flyspray

yetist:
Please review, see [FS#2107](https://bugs.flyspray.org/2107).


yetist:
Because this version of the code can not automatically synchronize login status, still need to make changes, so close it now.

PR #548, opened 2016-03-29
Bugfix in efforttracking - empty efforts

rkCSD:
It was possible to add empty efforts. The frontend displayed this efforts as tracked-effort and it wasn't possible to stop the tracking.


rkCSD:
Corrected last commit. Under some circumstances, effort-tracking wasn't possible

peterdd:
Thx for finding the bug.

* Please seperate translation from bugfix pull requests.
* The translation for invalidprogress and invalidpriority differs from invalidstatus. progress, priority and severity are within fixed ranges within Flyspray, while status can differ from project to project. so no need for the 'move to project' part here.

Existing multiple invalid shown "Tracking in Progress" entries of a user can be removed by the cancelTracking() function for the user. I will do it in another patch.

peterdd:
alternative fix implemented.

PR #549, opened 2016-03-31
Bugfix in displaying avatars

rkCSD:
If a user is in the "assigned-to"-list, his avatar displayed smaller than other avatars. This commit repair a malfunction in the avatar-cache.

### DIFF
--- a/includes/class.tpl.php
+++ b/includes/class.tpl.php
@@ -312,32 +312,38 @@ function tpl_userlinkavatar($uid, $size, $class='', $style='')
 	static $avacache=array();
 
 	if($uid>0 && (empty($avacache[$uid]) || !isset($avacache[$uid][$size]))){
-		$sql = $db->Query('SELECT user_name, real_name, email_address, profile_image FROM {users} WHERE user_id = ?', array(intval($uid)));
-		if ($sql && $db->countRows($sql)) {
-			list($uname, $rname, $email, $profile_image) = $db->fetchRow($sql);
-		} else {
-			return;
+		if (!isset($avacache[$uid]['uname'])) {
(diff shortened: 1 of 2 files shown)


peterdd:
Verified.

I wrote the simple $avcache array mainly to save sql queries for one page request.
If a page needs 2 different sizes (assigned list and comments for example) of the same useravatar the patch would still fire 2 sql queries.
The sizes are set at the moment as width and height parameter in html.

We could replace it by css class rules only and keep avcache simple, not?

The only difference is that gravatars are loaded from gravatar site really in different sizes. (when gravatars are enabled). For this a subarray is acceptable...

rkCSD:
Ok, I updated my pull-request. Now, only one SQL-Query per user will be fired even gravatar is set or not.

peterdd:
Mmh, I see still a problem. Not at current source, but maybe in future:

This example:

$first=tpl_userlinkavatar(10, 50, 'class1', 'border:5px solid red');
$second= tpl_userlinkavatar(10, 50, 'class2', 'border:5px solid blue');
echo $first.$second;  # would print 2 red borders, should print one red and one blue border

Maybe a sprintf('<a class="%d" ...  </a>', $class, $style )  would do the job too instead of building a $size sub array.

peterdd:
Well, it should never be required to pass a class or inline style to the tpl_userlinkavatar() function.

In current FS source the only calls to tpl_userlinkavatar() with a class parameter is tpl_userlinkavatar($user->id, $fs->prefs['max_avatar_size'], 'av_comment');

... and even that `av_comment` class is not needed there, because it can be addressed with other css rules like `.comment_avatar > a`

.. so maybe lets add a note that class and style parameter should not be used when calling tpl_userlinkavatar()

Explicit setting the 50 px size in the css may collide with the max_avatar_size preference setting.

PR #550, opened 2016-03-31
Bugfix with quickediting estimate_efforts

rkCSD:
This little bugfix repairs the quickediting for estimate_efforts. It's kind a little bit dirty, I think, because the quickedit-functions doesn't have a "do"-function. The edit of the header-file is needed for the quickedit.php on line 46-48. This lines need a correct project-Object. Previously the default-project was choosen there...

### DIFF
--- a/header.php
+++ b/header.php
@@ -77,7 +77,7 @@ text-decoration: none;
 }
 
 // Any "do" mode that accepts a task_id field should be added here.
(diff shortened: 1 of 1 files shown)


peterdd:
Thx for finding the bug.

#553 enhances #550. 
comments on #553 welcome.

peterdd:
see #553

PR #552, opened 2016-04-06
Update class.backend.php

PatrykCysarz:
Switched lines to fix bug - when searching for tasks assigned to user with ID = 4 it also returned tasks for user which had 4 in his username.

### DIFF
--- a/includes/class.backend.php
+++ b/includes/class.backend.php
@@ -1618,12 +1618,12 @@ LEFT JOIN {cache} cache ON t.task_id=cache.topic AND cache.type=\'task\' ';
                         $temp .= ' ass.user_id is NULL OR';
                     } else {
                         foreach ($db_key as $singleDBKey) {
-                            if (strpos($singleDBKey, '_name') !== false) {
-                                $temp .= ' ' . $singleDBKey . " $LIKEOP ? OR";
-                                $sql_params[] = '%' . $val . '%';
-                            } elseif (is_numeric($val)) {
+                            if (is_numeric($val)) {
                                 $temp .= ' ' . $singleDBKey . ' = ? OR';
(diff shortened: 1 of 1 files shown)


peterdd:
I do not see this as bug, more as a feature to use the sql like operator for a number search in user_id, user_name and real_name.

The autocompleter helps you set the right username to the field.

But I see a little bit inconsistence between the field and result list. The autocompleter doesn't show a user of that user_id when only a number passed to the autocompleter. But the task search shows then assignees with the number in user_id.

Maybe a radio group switch beside the field (e.g. labeled "default with like operator", "starting with", "exact username", "user_id") would solve it, but an implementation would probably the advanced search form even more clutter, not? Ok, with a bit of CSS or JS that radio group could be hidden with a on/off toggle so only a small fonticon as a hint for more sophisticated filtering option, that would be a new feature.

You can write a feature request on bugs.flyspray.org

peterdd:
* "default with like operator": username LIKE '%?%' or realname LIKE '%?%' or user_id=?
* "starting with": username LIKE '?%' or  realname LIKE '?%'
* "exact username": username LIKE ? or realname LIKE ?
* "user_id": user_id=?

PatrykCysarz:
I think you got me wrong. Let me explain you this by example. Let's say we have two users, patryk(id = 1) and peterdd1(id = 2). When peterdd1 searches for his tickets, he will also see tickets of patryk, because he has his ID number in username.

peterdd:
Ah, I understand now. You mean the 'my assigned task'-button. Mmh, in fact it is just another search query...

I opened an new task: bugs.flyspray.org/2121

Just switching like in your patch would fix the my assigned task-button, but would create a bug in the advabced search too.

PR #557, opened 2016-04-20
Another little fix in url_rewriting

rkCSD:
The multi-task-view can now accessed again


peterdd:
if you have projects >= projectid 10

PR #563, opened 2016-06-01
Updated zh-TW

CasparChou:


peterdd:
Hello @CasparChou,

as a dude who can't read chinese I'm asking what are the main differences compared to zh_cn ?

I just ask because another dialect of a language means also a bit more maintainence.

CasparChou:
Maybe they seen very close , but as a Taiwanese or Hongkongese , they prefer to switch language to Traditional Chinese ( because Simplified Chinese(zh-China) is hard to read , and their phrase is something strange  ) , so I hope I can make this version to help yours product more complete .
Thank you .

peterdd:
Ok, thanks for clarification.

PR #566, opened 2016-06-10
Allow admin to change own password

thess:
Without this change, admin can not change own password. Gets "no old password given" error. Change 6f52db86d64f78e515865de29ccf683dc4f409ff removed old password text box from admin "edit user" page. However, check for old-password required did not correct check for 'is_admin'.

ref: FS#2130


peterdd:
I applied a slightly different patch. But thx for the finding.

PR #586, opened 2016-09-16
Optimization of cryptPassword method

landrok:
- A simple Calculate Only Once optimization
- Usage of single quote
- Fix some indents


landrok:
Close this patch because indents are worst like that. Will pull again the optimization patch as a single new one.

PR #609, opened 2016-12-13
includes/class.backend: show max_date also when there is no comment.

kinimodmeyer:
this is a critical bug, you can´t see the last change date in the ticketlist if there is no user comment.
http://dev.mysql.com/doc/refman/5.7/en/comparison-operators.html#function_greatest
GREATEST() returns NULL if any argument is NULL.


peterdd:
I now merged #592, which is the same patch.

Probably related: https://bugs.flyspray.org/2011

kinimodmeyer:
ok ...

PR #611, opened 2016-12-27
Image transparency color, tab switching class

CaraesNaur:
These are a couple small things I've fixed in my local install.

Uploading an avatar that contains black masks that and creates ugly visual holes, so I switched the mask color to a more traditional one that is less likely to appear in images.

The tab switch class change here is more explicit.  I'm running a custom theme that is a complete copy of CleanFS, where I've used FontAwesome to put icons on the tabs by adding `fa-*` classes to the tab links.  There are several changes in this theme which I hope to backport into CleanFS soon as pull requests.

### DIFF
--- a/includes/modify.inc.php
+++ b/includes/modify.inc.php
@@ -40,9 +40,9 @@ function resizeImage($file, $max_x, $max_y, $forcePng = false)
 	$newHeight = $height * $scale;
 
 	$img = imagecreatefromstring(file_get_contents($src));
-	$black = imagecolorallocate($img, 0, 0, 0);
+	$transparent = imagecolorallocate($img, 255, 0, 255);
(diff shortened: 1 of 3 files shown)


peterdd:
Hello CaraesNaur,

thx for your help. Please do one pull request per problem you solve.

This makes decision to accept a pull request easier and each problem can be discussed separatly.

CaraesNaur:
I've used git for a long time, but this is my first foray into pull
requests.  I didn't realize until after my third commit that they're
branch-based.  Would you like me to retract it, split my changes into
branches, and resubmit?

peterdd:
Yeah, would be the easiest for me and others.

CaraesNaur:
Closing this pull request, will resubmit individually.

PR #613, opened 2016-12-28
Uploaded image transparency

CaraesNaur:
Avatars that contain black have it made transparent, which is visually unappealing.  Switched transparency color to fuschia (#f0f).

### DIFF
--- a/includes/modify.inc.php
+++ b/includes/modify.inc.php
@@ -40,9 +40,9 @@ function resizeImage($file, $max_x, $max_y, $forcePng = false)
 	$newHeight = $height * $scale;
 
 	$img = imagecreatefromstring(file_get_contents($src));
-	$black = imagecolorallocate($img, 0, 0, 0);
+	$transparent = imagecolorallocate($img, 255, 0, 255);
 	$resizedImage = imageCreateTrueColor($newWidth, $newHeight);
-	imagecolortransparent($resizedImage, $black);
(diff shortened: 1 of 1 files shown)


peterdd:
I found this on http://php.net/manual/en/function.imagecolortransparent.php#28035

> 
> I have found that ImageColorTransparent() works on true color images if and only if the transparent color is black.
> (PHP 4.2.3/GD 2.0)

Well, comment 14 years old and maybe this why black was choosen in Flyspray (guessing). What is the status in 2017?
Ideally the resizeImage() of Flyspray function detects the correct transparency values automatically for any type of uploaded gif and png.

CaraesNaur:
I'm pretty sure that comment predates when PHP forked GD, so that issue may be fixed.

I agree that detecting an alpha channel first is a better solution, along with preserving any existing transparency instead of creating a new channel based on an (ultimately) arbitrary color.

I'll continue working on the better long term solution.

peterdd:
Just if someone is interested in working on this: GD seems to be often a candit for bugs/fixes between php versions, for instance http://www.php.net/ChangeLog-7.php#7.0.11 https://bugs.php.net/bug.php?id=66005
So maybe figure out which works when for our current supported PHP-range php5.4-php7.1 (edge cases php5.3.*/php7.2.x+)

peterdd:
inactivity of pull request. Do a new PR with a long term solution if problem still exists.

PR #614, opened 2016-12-28
Active tab class switching

CaraesNaur:
Changed add/remove of active class on tabs to be more explicit, which is safer for themes that may have added other classes to tabs (ie, FontAwesome).

### DIFF
--- a/js/tabs.js
+++ b/js/tabs.js
@@ -49,8 +49,8 @@ function showTabById(tabid, noEval) { // {{{
       var links = submenu.getElementsByTagName('a');
       for (i=0; i<links.length; i++) {
         if (links[i].href.match('^.*#'+tabid+'$')) {
-          links[i].className = 'active';
-        } else { links[i].className = ''; }
+          links[i].className += ' active';
+        } else { links[i].className = links[i].className.replace(/ active\b/, ''); }
       }
     }
   }


peterdd:
I wish one day I can replace the javascript tab switcher to a pure css3 tab switching that also adapts to different screen sizes (responsive/adaptive layout). E.g. show task comments on wide screens independent of the other tabs.

Spreading fontawesome classes (fa fa-lg ..) into the HTML theme templates is not ideal. I also feel a bit guilty about it within Flyspray 1.0.

You could add icons to the tab labels also without adding a class, e.g. in your custom_***.css:

`#submenu #notifytab a::before {
    content: "\f007";
    font-family: FontAwesome;
    padding-right: 4px;
}`

CaraesNaur:
Frankly, I'm surprised by the amount of vanilla JS in FlySpray see even when Prototype is in play.

Pure CSS widgets is somewhat of an ideal, but not universally achievable.  I would consider responsive layout to be a separate goal (and it opens a nasty can of worms).

FontAwesome is something a project needs to decide whether to fully embrace.  It's a far better solution than the early days of icon fonts where the icon glyphs would occupy the ASCII page in the font.  Buttons (along with other kinds of actionable elements) gain usability when a glyph is present.

That CSS line is more or less what I did in my theme, however the JS that twiddles the active tab completely wipes out the element's className, which is a potential problem for any custom CSS on the tabs, not just FontAwesome.

peterdd:
What I wanted to tell is that that a-tags on the tab labels never need an extra css class because someone can address that tags with for instance `#submenu #notifytab a` and avoid any further complexity by the regexp. (I am not familiar with reg. expression in javascript and wanted avoid future surprises)

Or do you have a situation where that is needed?

CaraesNaur:
FontAwseome is only incidental to this commit.  The purpose of it was to make removal of the active class a perfect reversion of adding it.

When I was adding FontAwesome (incidental except that I was using classes to do it) to the tabs, switching between them was removing all the classes when a tab became non-active.  I didn't understand why until I examined the tab switch code.  Any additional classes, regardless of purpose, are no longer wiped out.

I am going to alter my FontAwesome usage to avoid peppering additional classes in the markup.  But that doesn't change the fact that the way the active class previously got removed was too aggressive.

peterdd:
Yea, the active insert/remove as it is/was too aggressive.

I see the FontAwesome usage as intermediate technology within Flyspray. It is 'ok' at the moment and enabled us to quickly add some ready applike icons. Could be finally replaced by SVG in future so I want avoid spreading fa-classes in templates if possible.

peterdd:
A little bug with this commit: click 2 (or more) times on a tab label, click on another tab label--> 2 tabs are now 'active'. 

But I still wish to make tabs behavior works CSS only, at least for the task details page.
So that tabs.js is not needed for this page type.

CaraesNaur:
Confirmed.  Both sequential clicks and doubleclick events on the active tab trigger this bug.
A CSS-only solution will possibly impact browser version requirements for IE and possibly Safari, what are the those targets?

PR #617, opened 2017-01-10
Remove some dead JS code

landrok:
The following functions are no longer used:
- adduserselect
- setUpSearchBox
- toggleSearchBox


peterdd:
yup, found adduserselect usage in 0.9.9.2, not in 0.9.9.7 so should be ok to remove.
setUpSearchBox toggleSearchBox replaced by pure css toggle.

PR #618, opened 2017-01-10
Replace direct access to input var by using a GPC method

landrok:
- Usage of GPC post method
- Remove unnecessary else + move OK comment for clarity purpose

### DIFF
--- a/includes/modify.inc.php
+++ b/includes/modify.inc.php
@@ -952,15 +952,14 @@ switch ($action = Req::val('action'))
             break;
         }
 		
-		if(!($user->perms('is_admin')) && $fs->prefs['captcha_securimage']){
+		if (!($user->perms('is_admin')) && $fs->prefs['captcha_securimage']) {
 			$image = new Securimage(); 
-			if ($image->check($_POST['captcha_code']) == true) {
-				# ok  
-			} else {
+			if (!$image->check(Post::val('captcha_code'))) {
 				# wrong code
 				Flyspray::show_error(L('captchaerror'));
 				break;
 			}
+			# captcha is OK
 		}
 
         if (!Post::val('user_name') || !Post::val('real_name') || !Post::val('email_address'))


peterdd:
Hi landrok, could be a totally dumb question, but what is the benefit of using Post:val() instead of $_POST[]?
I know I wrote the captcha integration in a hurry to stop spambots, but whats so bad with passing $_POST[] here? Better 'testability'?

This is also a general question, sometimes that GPC layer in Flyspray looks so redundant to me.

landrok:
Yes there is. A simple use case is when `$_POST` array key is not set. It would raise a notice. (For vicious bots).
Furthermore, I had a look at the check method, multidimensional arrays (ie. `$_POST['captcha_code']['my_second_dimension']`) are well skipped. (For very vicious bots)

Our GPC layer is a good abstract - could be improved. You're true, better testability is one of the good aspects to consider with this type of abstraction.

peterdd:
What do you mean with 'are well skipped'? That the captcha can be bypassed?
I tried with an passing as post-array (captcha_code['blabla'], got only a warning, no bypass the check().

It spits a warning Warning: strtolower() expects parameter 1 to be string, array given in /***/dapphp/securimage/securimage.php on line 2358

Same with using Post::val() 
Will put a Post::inAlnum() before passing to $image->check()

landrok:
Nice catch! There is a mistake in their "validate" method 
https://github.com/dapphp/securimage/blob/master/securimage.php#L2332
So, I will pull a little patch, less restrictive than inAlnum().

peterdd:
I think we should stick to the default character set and keep it simple as possible:
https://www.phpcaptcha.org/documentation/customizing-securimage/#charset

which is a subset of [a-zA-Z0-9] as the documentation says.

landrok:
So, https://github.com/Flyspray/flyspray/commit/af1b61f50e59132a8276036b1158a00646b731dd is great.

PR #619, opened 2017-01-10
Fix install typos

landrok:
- Fix PHP install. @todo To replace by php7 in the near future
- F5 / CTRL+F5 @see http://stackoverflow.com/questions/385367/what-requests-do-browsers-f5-and-ctrl-f5-refreshes-generate, depending on which component is managing the cache (browser, proxy, server)

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,14 +35,14 @@ http://flyspray.org/manual/install
 4. Make sure flyspray.conf.php is writeable by the webserver.
 5. Open http://yourflyspray/setup/ in your webbrowser. It detects the existing installation and you can follow the upgrade steps.
 
-6. Note: Do not forget to press F5 after the upgrade in web browser to reload also cached css-files to see effects of updated CSS-files. They are cached by default for 14 days in the webbrowser.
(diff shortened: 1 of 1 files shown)


peterdd:
CTRL+F5 notice is ok and will merge if you adapt the pull request.

name of php->php5 change not. Maybe extra description about which executable to use in a shared web hosting or managed server environment. In my web hostings is see many different names for different available php versions:

php php4 php4.4-cli php5.2 php5.4 php5.5 php6 (for a php5.4.45 cgi version, lol) php4.4 php5 php5.2-cli php5.4-cli php5.5-cli php-5.2 php-5.3 php-5.6 php-7.0 php-cgi php-cgi-5.2 php-cgi-5.3 php-cgi-5.6 php-cgi-7.0

Besides from the antique php4 versions, the user must often decide to choose the best supported choice there (combined with setting which php version for apache should be used). Web hostings provide that for (sub)domains or file directories.

So lets keep the generic 'php' string there and better add an explanation.

peterdd:
sorry, haven't read the pull request exactly..

PR #621, opened 2017-01-11
Bigimages lightbox

peterdd:
This modifies the Lightbox in a way that attachment images that are wider than the browser window fit into the window.


Thanatermesis:
I wonder why this fix is not merged yet

PR #622, opened 2017-01-12
Some additions and a typo fix

akisc:


peterdd:
Hi akisc,

please do not mix addressing different issues within one pull request.
I cannot merge this at it is, but would like to merge parts of it.

For instance I regret a bit adding the class field to the tag-dbtable. At least at the current status of Flypspray, because undere some condition the classes got spread between tags at display in the task list and other edge cases. So I will not encourage extending that idea also to other tables like tasktypes - not at that moment.

Instead stick with writing the css for each task type to your custom_*.css

peterdd:
Typo fix is merged now. 

h2 is a bit more logic than h3-tag at that position you suggested. Well, theoretical it should even be h1.
However is see it is inconsistent with the other admin areas, which currently use h3-tags for that.

See also my comment in CleanFS/templates/common.list.tpl :
# TODO: should be h2 tag, h1 tag for page type title, and project/flyspray title not a h1-tag in the header.

akisc:
Hi peterdd,

As for the classes, the initial idea was to be able to define a color in types such as red letters in the "Bug Report". But when I saw the code found that uses classes prefixed with the word "typ" + the ID of the database record, but due because that it is not stable, I thought instead of having a standard name, to allow the user to define a name in class. All these classes are defined in custom_*.css Finally if the user does not define the class name, then they will run as it was.

PS 1: Sorry for the mixing different issues, it's my first time contributing in a public project. I with my team we work differently. If there is a standard to follow in please let me know because writing a template based on CleanFS.

PS 2: You did well not proceed with Pull Request from my mistake I forgot to write the condition to check if the class is defined to show it.

PR #623, opened 2017-01-19
Fixed theme part

akisc:
class.tpl.php line: 85 (FIXME: Shouldn't have to do this but there is a bug somewhere cause theme to sometimes come in as empty)

This was because there were new instances of FSTpl class in code instead of the initial was created.

To avoid "empty" theme use the initial constructor of Tpl (FSTpl) by accessing the global variable
$page = $GLOBALS['page']

### DIFF
--- a/includes/class.tpl.php
+++ b/includes/class.tpl.php
@@ -81,8 +81,9 @@ class Tpl
             return;
         }
 
-        // theming part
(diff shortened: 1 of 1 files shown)


peterdd:
@akisc I merged @landrok 's pull request now. We can improve from there then altogether.

peterdd:
#539 addressed the same issue.

PR #625, opened 2017-01-24
fix FS#2330

peterdd:
and build the appropriate .csv filename also for the 'all projects' tasklist when doing csv export from a task search.

### DIFF
--- a/scripts/index.php
+++ b/scripts/index.php
@@ -427,7 +427,7 @@ function export_task_list()
 
         usort($tasks, "do_cmp");
 
-        $outfile = str_replace(' ', '_', $tasks[0]['project_title']).'_'.date("Y-m-d").'.csv';
+        $outfile = str_replace(' ', '_', $proj->prefs['project_title']).'_'.date("Y-m-d").'.csv';
 
         #header('Content-Type: application/csv');
         header('Content-Type: text/csv');


landrok:
Some characters can break the filename. `"` or `\` for example.
Furthermore, multiple consecutive spaces or replacements are giving multiple consecutive underscores.
Maybe we should use `preg_replace` instead of  `str_replace` in order to have an explicit list of allowed characters (whitelisting).

```php
// digits and characters only
$outfile = sprintf(
  '%s_%s.csv',
  preg_replace('/[^\d*|^\w*]+/', '_', $proj->prefs['project_title']),
  date("Y-m-d")
);
```

peterdd:
test_me

At least a project named _test"me_ works with " in csv export..

peterdd:
test2

peterdd:
Previous current Firefox, same testprojectname looks a bit different on IE (Edge) ...
test2_ie

landrok:
Yes, in these cases, the renaming process is delegated to the browser.
If you look at the HTTP headers, for the `test"me` case, you will have something like 
`Content-Disposition :"attachment; filename=test"me_2017-01-24.csv"`

For the second case, as far as I can see, it's an encoding problem. I don't have any IE in my current location, I can't really help more.

peterdd:
Well, not as easy as thought: http://stackoverflow.com/questions/93551/how-to-encode-the-filename-parameter-of-content-disposition-header-in-http

peterdd:
I will open another task for that filename filtering on bugs.flyspray.org ...

landrok:
For IE, can you try this:
```php
$outfile = sprintf(
  '%s_%s.csv',
  utf8_decode(preg_replace('/[^\d*|^\w*]+/', '_', $proj->prefs['project_title'])),
  date("Y-m-d")
);
```

peterdd:
Any 'umlaut' stripped away before utf8_decode...

I think we need something like

```header('Content-Disposition: attachment; filename="'.$outfileascii.'"; filename*=utf-8\'\''.$outfileutf8);```

see RFC5987

landrok:
The following seems good:
`header('Content-Disposition: attachment; filename=*=utf-8\'\''.$outfile);`

peterdd:
played a bit, now IE-edge saves with uft8 chars, but firefox and chrome want save the ascii filename...

PR #626, opened 2017-01-28
fix a PHP7.1 notice: A non well formed numeric value encountered ..

yagu0:
$val initially contains either an integer, or a string like 128M or 512K.
Therefore, to obtain the value in octets we need to

  1) get the last character and store it
  2) trim it from $val to get a correct number
  3) multiply $val by the amount obtained through the switch control

That's what the patch does.

### DIFF
--- a/includes/class.flyspray.php
+++ b/includes/class.flyspray.php
@@ -95,14 +95,14 @@ class Flyspray
                 continue;
             }
 
-            $val = trim($val);
             $last = strtolower($val{strlen($val)-1});
+            $val = trim($val, 'gGmMkK');
             switch ($last) {
                 // The 'G' modifier is available since PHP 5.1.0
                 case 'g':
-                    $val *= 1024;
+                    $val *= 1024*1024*1024;
                 case 'm':
-                    $val *= 1024;
+                    $val *= 1024*1024;
                 case 'k':
                     $val *= 1024;
             }
--- a/includes/class.flyspray.php
+++ b/includes/class.flyspray.php
@@ -100,9 +100,9 @@ class Flyspray
             switch ($last) {
                 // The 'G' modifier is available since PHP 5.1.0
                 case 'g':
-                    $val *= 1024*1024*1024;
+                    $val *= 1024;
                 case 'm':
-                    $val *= 1024*1024;
+                    $val *= 1024;
                 case 'k':
                     $val *= 1024;
             }


peterdd:
Example? Existing code seems to give correct results, see 
also http://php.net/manual/en/control-structures.switch.php

yagu0:
Oops you are correct about the cascading cases, multiplications are OK :) I'll remove them.

However the other lines were *not* OK at least on last release (I had PHP errors saying in short that "at line 105 $val is not a valid number", which were fixed by using the above lines). As confirmed by some traces (I just echoed $val) and logically explained in the pull request :)

Procedure to reproduce the PHP error : download and install last release, and then just open main welcome page.

yagu0:
Example:

Let's assume you have post_max_size defined at "5M" in php.ini.

With the current code, $val will initially contain the string "2M".
Then:
$val = trim($val); <-- does not perform anything
$last = strtolower($val{strlen($val)-1}); <-- $last contains 'm'
case 'm': $val *= 1024; <-- problem here

With the patched code:
$last = strtolower($val{strlen($val)-1}); <-- last contains 'm'
$val = trim($val, 'gGmMkK'); <-- $val contains 5
case 'm': $val *= 1024; <-- everything fine, 5*1024

peterdd:
Can't reproduce the error message yet. Which PHP version and error reporting levels?

yagu0:
Steps:

git clone https://github.com/Flyspray/flyspray.git
cd flyspray/
git checkout v1.0-rc4
php -S localhost:8000
composer install

Then navigate on http://localhost:8000
PHP version : 7.1.1
All flags green on http://localhost:8000/setup/index.php except MySQL (I use PostgreSQL)
Then provide DB credentials, choose admin name/pwd, and finally click "view site" :

Notice: A non well formed numeric value encountered in /home/auder/flyspray/includes/class.flyspray.php on line 105 Notice: A non well formed numeric value encountered in /home/auder/flyspray/includes/class.flyspray.php on line 105 Notice: A non well formed numeric value encountered in /home/auder/flyspray/includes/class.flyspray.php on line 105

(Same errors after removing setup folder). If you put some traces you'll see the execution I mentioned in the example above.

Error level ( using this function http://stackoverflow.com/a/9167563 )
E_USER_DEPRECATED,E_RECOVERABLE_ERROR,E_USER_NOTICE,E_USER_WARNING,E_USER_ERROR,E_COMPILE_WARNING,E_COMPILE_ERROR,E_CORE_WARNING,E_CORE_ERROR,E_NOTICE,E_PARSE,E_WARNING,E_ERROR

Anyway whatever the error level, just tracing $val should display something wrong. The current code cannot do the correct computation (because trim($val) doesn't trim anything: there are no whitespaces to trim). I hope you will reproduce the error and fix it :)

peterdd:
It seems to be a change of PHP's behavior starting with PHP7.1 :-(
See https://wiki.php.net/rfc/invalid_strings_in_arithmetic

peterdd:
Why removed the first trim()? Is it impossible to have `'  5M   '` as $val? Ask because someone must had a reason to add the first trim() there..

yagu0:
My guess would be that the first trim is not necessary (confirmed by a php -r "echo('|'.ini_get('post_max_size').'|');" on my system after adding white spaces around post_max_size in the .ini file); but it doesn't hurt to let it for the moment (maybe required in some PHP versions).

I've read the link you indicate, it's interesting, I had no idea PHP accepted $numberOfApples = "10 apples" + "5 pears"; ^^ So I understand better now why the current code worked (and still work but with warnings).

PR #627, opened 2017-02-15
FS#2309 - PHP noticed displayed on default "All Projects" page.

Arthmoor:
Code to fix 2309. It's been tested for some time now and no adverse effects have been noted.

### DIFF
--- a/includes/class.flyspray.php
+++ b/includes/class.flyspray.php
@@ -642,19 +642,9 @@ class Flyspray
   public static function cryptPassword($password)
   {
 	global $conf;
-	$pwcrypt = strtolower($conf['general']['passwdcrypt']);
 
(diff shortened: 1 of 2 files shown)


peterdd:
Ignoring passwdcrypt setting in cryptPassword() is breaking compatibility. Not between rc-releases.

Arthmoor:
That may be, but leaving insecure, easily cracked passwords floating around, doesn't really make a lot of sense either. All of the schemes the compatibility setting was being used for in the code are all for unsalted passwords on encryption methods that are considered completely broken today.

peterdd:
Added your php notice fix now.

The other issue with upgrading old passwordhashes to current recommended algorithms must be done without breaking compatibility.

PR #630, opened 2017-03-22
Active tab class switching improvement

CaraesNaur:
When switching tabs, remove all instances of the active class from all tab links' className before appending ' active' to the target link.  Solves the issue introduced in pull request #614: the active class was getting added multiple times, but only the final instance got removed.

FYI: 'CSS-only' tab switching doesn't seem possible in the strictest sense.  There are two popular approaches considered 'CSS-only':

- Replace tab links with radio input+label pairs and use CSS selectors to show the correct panel; all the implementations I've seen rely on ~ selectors, which require the radios and panels to be siblings (uglier DOM structure).  It may be possible to use more complex attribute selectors instead, and maintain the current tabs DOM structure.
- Use the target URL fragment to indicate the active tab; most implementations rely on JS to prevent the window from jumping up and down.

### DIFF
--- a/js/tabs.js
+++ b/js/tabs.js
@@ -49,8 +49,8 @@ function showTabById(tabid, noEval) { // {{{
       var links = submenu.getElementsByTagName('a');
       for (i=0; i<links.length; i++) {
         if (links[i].href.match('^.*#'+tabid+'$')) {
-          links[i].className += ' active';
-        } else { links[i].className = links[i].className.replace(/ active\b/, ''); }
+          links[i].className = links[i].className.replace(/ active\b/g, '') + ' active';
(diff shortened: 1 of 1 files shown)


peterdd:
> Replace tab links with radio input ...
I would prefer using checkboxes and some css logic is possible. (maybe dynamic generate this on/off logic css part serverside depending on user permission for available shown task details tabs)

For instance enables a tab displayed separatly depending on window width. An example would would be 'always show content of the comments tab if the screen(media query) is wide enough'. Like a split screen where the comments are displayed on the right side beside tast description.

CaraesNaur:
That is an order of magnitude more complex, and would transform tabs into something else.

PR #635, opened 2017-05-11
Updated export of task list to csv file

theunderstated:
Initially I wanted to change which columns could be exported from Flyspray which would have been easy by simply modifying the code but then I decided to try and make the columns selectable from within the software. This was a bit more work than I first thought but I have managed get most of what I wanted to get done.

I've added an extra setting to the users settings which is used to hold the names of the task columns to be exported. So, I've also had to make modifications to the database to allow for this extra field. I've also added an extra field to the users profile page which utilises the dual list display to allow the user to select which columns they want exporting when the click 'export tasks' on the main form. I haven't limited the column names according to the users permissions though this would not be too difficult to do at some point.

I have played with a number of files while trying to work out what needed to go where but the only files that have changed are:-

- includes\class.backend.php
- includes\modify.inc.php
- scripts\index.php
- setup\upgrade\1.0\flyspray-install.php
- setup\upgrade\1.0\upgrade.info
- setup\upgrade\1.0\upgrade.xml
- themes\CleanFS\effects.css
- themes\CleanFS\theme.css
- themes\CleanFS\templates\common.profile.tpl


peterdd:
Please solve only one problem per pull request and keep changes for a pull request as small as possible for easier review.

For instance:

- The change from default 50 to 250 in scripts/index.php has nothing to do with the topic of pull request, default limiting to 50 for anon guests was done for uncached description dokuwiki/geshi.
- fixing whitespace to tabs in areas that are not required to change for pull request.
- adding effect.css not required for pull request.
- view_estimated_effort field permission not handled as in original

Whats wrong with using fputcsv() instead of own workarounds?

theunderstated:
Hello Peter,

Sorry for the extra work that I've caused you. To answer your questions.

The number items per page being set at 250 is my default setting and was left in unintentionally.

The white space being converted to tabs is something that I tend to do by default when indenting but forgot not to do this here.

Effects.css only because I used the effects when adding extra options to the users preferences.

View_estimated_effort I didn't think I was doing anything special with. My use for it is to be able to export it's value into the .csv

The reason for not using fputcsv as I did originally and reduced everything down to a small number of lines is that when I exported data it got confused by some of the contents in some the fields. I intend on passing the exported csv files to a script file for processing so it became far less problematic when I decided for format the fields myself. This is made the script much less likely to encounter problems which matters a lot when processing many tasks. So, for me I needed it to be robust over being convenient.

At the moment the feature works well for me and I know that there are certain bits which are unfinished like the labels/descriptions displayed haven't been entered correctly in the way that the other labels have been setup. However, feel free to change what you wish. I did expect aspects to be changed as my update doesn't currently deal with user permissions and I think you are better placed than I am to decide what a user should be able and should not be able to export.  

Finally, I'd like to say thank you very much for all the time and good work that you have put in keeping this project going

peterdd:
not forgotten, https://bugs.flyspray.org/?string=csv&search_in_details=1

theunderstated:
Apologies for the late reply. 

The issue with the export function not working properly for fields that have double quotes embedded within them I initially got around by not using the 'fputcsv' function. Instead, I manually formatted each exported field which was a less elegant way of doing it but a more robust and  reliable method that gave predictable results.

The fields to be exported are selected in the preferences when logged in as the administrator.

PR #639, opened 2017-09-18
fix FS#2218

peterdd:
The increasing of field attachment.file_type size from 50 to 255 between 0.9.9.2 to 0.9.9.4 and upward was missing in the upgrade.xml, only in flyspray-install.xml . So a Flyspray installation that was first installed <=0.9.9.2 and then upgraded later never got this change in the database.


peterdd:
https://tools.ietf.org/html/rfc6838#section-4.2

PR #646, opened 2017-10-08
index.php: workaround PHP bug 43225

jow-:
PHP's implementation of fputcsv() fails to properly escape CSV quote
characters preceeded by a backslash. An input string like `test \" test`
is converted to `"test \" test"` instead of the expected `"test \"" test"`
result.

Due to this issue, it is impossible to properly parse the generated
task CSV export list.

Work around the problem by utilizing a custom CSV formatting function
instead which properly escapes quote characters.

Signed-off-by: Jo-Philipp Wich <jo@mein.io>

### DIFF
--- a/scripts/index.php
+++ b/scripts/index.php
@@ -369,6 +369,25 @@ function do_cmp($a, $b)
    return ($a[ $orderby ] > $b[ $orderby ]) ? -1 : 1;
 }
 
+function my_fputcsv($handle, $fields)
+{
+  $out = array();
+
+  foreach ($fields as $field) {
+    if (empty($field)) {
+      $out[] = '';
+    }
+    elseif (preg_match('/^\d+(\.\d+)?$/', $field)) {
+      $out[] = $field;
+    }
+    else {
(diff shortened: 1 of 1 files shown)


jow-:
This PR is meant to draw some attention to the CSV formatting problems when exporting the task list.
We're successfully using this fix at https://bugs.lede-project.org/ since we do have some external scripts and utilities using the task list export do cross-reference things.

When you decide to merge that to master, you likely want to move `my_fputcsv()` into a central place and use it everywhere to avoid incensistencies among different exports.

jow-:
This will also fix FS#2338.

peterdd:
What about this https://bugs.php.net/bug.php?id=43225#1442378006 

This is not working (also a php bug I think)
```
fputcsv($output, $row, ',', '"', '"');
```

This is better:
```
fputcsv($output, $row, ',', '"', "\0");
```

Only php5.5.4+ ..


Also adodb has some csv related stuff, like rs2csv(), don't know if reliable.

PR #647, opened 2017-10-14
Error in HTTPClient.php on install/upgrade

bepixeld:
I get Error on install/upgrade
PHP Fatal error:  Cannot redeclare class DokuHTTPClient in [flyspray/plugins/dokuwiki/inc/HTTPClient.php](https://github.com/Flyspray/flyspray/blob/master/plugins/dokuwiki/inc/HTTPClient.php) on line 20
If I change the order of the classes, everything is working.
See also [Flyspray FS#2088](https://bugs.flyspray.org/index.php?do=details&task_id=2088)
with the same problem.

System: Apache2.4, PHP5.6

### DIFF
--- a/plugins/dokuwiki/inc/HTTPClient.php
+++ b/plugins/dokuwiki/inc/HTTPClient.php
@@ -13,33 +13,6 @@ define('HTTP_NL',"\r\n");
 
 
 /**
- * Adds DokuWiki specific configs to the HTTP client
- *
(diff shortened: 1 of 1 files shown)


peterdd:
Couldn't reproduce yet, maybe only under special circumstances. But lets follow PHP documentation at http://php.net/manual/en/language.oop5.inheritance.php

> If a class extends another, then the parent class must be declared before the child class structure.

PR #678, opened 2018-07-10
Upgrade pwhashes if possible

peterdd:
during login when user provided password


crrodriguez:
I'll leave my suggestion here too.. I strongly advise you not to provide configuration options for this..but read the old setting just to migrate to "PASSWORD_DEFAULT with no extra options" hashes.. that will evolve naturally in time to stronger/better algorithms. also make the database password field to 255 chars for futureproofing.

PR #688, opened 2019-01-04
Add "assigned" field to sub,dependency and blocking tasks to details …

poucz:
Adding assigned filed to tasklist detail.


peterdd:
Too many issues with this: shows wrong assignees on wrong subtasks, shows warnings in php7.2
Example: 
task1 without an assignee
  with
 subtask2 with 2 assigned users dev2, dev3
 subtask3 without an assignee
 subtask4 without an assignee

The helper function tpl_userlink($user_id) can be used instead of the implode().

The template calls $fs->getTaskDetails() in a foreach loop, this should be done by scripts/details.php, not the template file and the accessing $dependancy array there does match with the foreach loop in the template.

PR #715, opened 2019-03-19
It used the very-old logo

FrancescoGd:


peterdd:
The reason I did not merged is that I have some objections permanently accepting the 'newer' logo:
* the colors should play well with default CleanFS theme
* the less iconic design (We do need quadratic version for generating icons for favicon(32x32),  bigger for apple, android etc (https://realfavicongenerator.net/)
* the metaphor of the logo (a fly flying away? Mmh, so the fly can anytime come back instead of 'killing' it? It always reminds me of some advertising of an airline/travel agency.)

FrancescoGd:
Don't worry... I closed it because I changed my mind even though it irks me to have different logos in different pages :D
Anyway, if there isn't someone able to design a new logo/icon system, it may be worth looking for some free to use elements.

PR #724, opened 2019-07-13
Option to force url scheme (http|https)

martinlolek:


peterdd:
Mmh, I do not yet understand the problem your pull request tries to solve. If you want to force https for instance, why not just insert the full url as 'https://bugs.example.com'?

Please explain.

peterdd:
related: https://stackoverflow.com/questions/18008135/is-serverrequest-scheme-reliable

martinlolek:
I don't want to force https but just customize the user's URL request. If I put https that the user does not want to use, the user can get eg an error on an untrusted certificate. I added it as an option to the config so I don't mean to endanger anyone. If you want i change it to use is_ssl () function from wordpress from your link.

Neustradamus:
Any news on it?

Neustradamus:
Any news on it after more 2 years my previous ticket?

Neustradamus:
Any news on it after more 2 years my previous comment?

Neustradamus:
More one year after my previous comment, what is the PR situation?

Neustradamus:
@peterdd: It is possible to have news from you?

Thanks in advance.

PR #727, opened 2019-07-29
detect basic browser and platform at session start

peterdd:
to enable to show the user the key combination that works on his system


FrancescoGd:
I was just looking at this last friday 'cause it's not marked as "assigned" on bugs.flyspray... luckily I didn't do much beyond tinkering with browsecap! 🙂

PR #729, opened 2019-08-08
More intuitive layout

FrancescoGd:
Having the searchbox inbetween two buttons, and especially after the search button, is totally counter-intuitive (you're often lead to click the wrong export button instead of the search button)


peterdd:
Hello Francesco, sorry for delay.

I understand what you mean, but there are some things to consider:

- The translations of the 'searchthisproject' button differ slightly between languages. For english for instance it is 'Search this project for' 
So for english switching _search button_ and _search textfield_ would require rewording the button.

- The export button moved to the right with the current version by CSS. So the _search textfield_ is not sandwiched anymore on normal screen widths. This may not be the ideal solution, I know. (Maybe use html5 form attribute to put export button at end of result table to avoid confusion)

- I experimented a bit with showing some of the filter values of a search result on the same row like the _search button_ and _search textfield_ when the _extended search form_ is collapsed. In that case switching _search button_ and _search textfield_ would look like:

`___textfield___  [search button]  [icon1 x][icon2 x][filterval3 x][filterval4 x]`
and you are back to a similiar problem.

FrancescoGd:
Not a problem at all, I know it's summer for everyone (at least in this part of the globe 😉)
WRT the first point, I believe it'd still be better to have the input before the button and eventualy go through the hassle of rewording some translations. That's the de-facto standard norm in 99% of UIs and, on top of all, what people are accustomed to, and that's what should drive decisions when talking UI/UX.
I bet that if we had some metrics built in we would find lots of people mistakenly hitting the export button instead of the search button. 
From a usability point of view, even having **no search button at all** would be better than having it where users don't expect to find it.
I'm aware it's just a quick'n'dirty solution, which brings me to the rest of the bullet points which I believe are all part of a larger argument that I tried to point out in my comment on FS#1972 some months ago, if you saw it.

FrancescoGd:
Closed, not needed anymore

PR #730, opened 2019-08-28
feat: make ldap configurable

fch25804:
Changes:
* config parameter to set authentication method
* configurable LDAP authentication
* account provisionning from LDAP on first login
* fallback to native authentication if LDAP user not found

To configure LDAP authentication, set following parameters in config file
(flyspray.conf.php):

```
[general]
auth_method = 'ldap'

[ldap]
host = 'ldap.example.com'
port = '389'
version = '3'
base_dn = 'ou=users,dc=example,dc=com'
search_user = 'your_bind_dn'
search_pass = 'your_bind_password'
filter = 'uid=%USERNAME%'
field_name = 'cn'
field_email = 'mail'
```

### DIFF
--- a/includes/class.backend.php
+++ b/includes/class.backend.php
@@ -705,6 +705,9 @@ abstract class Backend
                 $newuser[0][$email] = array('recipient' => $email, 'lang' => $fs->prefs['lang_code']);
             }
 
+	    if(is_null($notify)) {
+		    $notify = new Notifications();
+	    }
             // Notify the appropriate users
 			if ($fs->prefs['notify_registration']) {
                 $notify->create(NOTIFY_NEW_USER, null,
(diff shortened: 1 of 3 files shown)


peterdd:
The Null Coalescing Operator ?? is available since PHP7 ( https://www.php.net/manual/en/language.operators.comparison.php ). Could you please adapt your PR so it doesn't break older PHP versions?

Further I see in apache log when fiddled with trying to test the pull request:

PHP Warning:  count(): Parameter must be an array or an object that implements Countable in /var/www/html/fsldap/includes/class.flyspray.php on line 815, referer: http://ldap.example.com/fsldap/

peterdd:
https://www.php.net/manual/en/function.ldap-connect.php says ldap_connect($host,$port) is deprecated and instead use 

`ldap_connect($ldap_uri);`

So we should use a `$conf['ldap']['uri']` variable instead of `$conf['host']` and `$conf['port']` , right?

fch25804:
True. Thanks.

PR #733, opened 2019-08-31
Another French translation update

trebmuh:
Hope that helps.
Olivier


lang/fr.php
```diff
@@ -254,7 +254,7 @@
 'systemvalues'            => 'Liste des valeurs globales',
 'projectvalues'           => 'Liste des valeurs spécifiques au projet',
 'remindthisuser'          => 'Destinataire du rappel : ',
-'thisoften'               => 'Période : ',
+'thisoften'               => 'Tou(te)s les : ',
```
peterdd: Is this french singular/plural or other grammar issue?
I prefer avoiding such constructs. Instead chose the correct wording on php output for instance if a number is known. Might need some change to the translation keywords too.
Is Periode not precise?

I see a few lines below:
hour ->heure
hours -> Heure(s)

Why is (s) here? Different meaning/context? I see some other language files like es, sl, mk, sr_lat, it, el, pt_br have similar workarounds. Is it due different usage of 'hours' string?

```
grep -rn "'hours'"
CleanFS/templates/newtask.tpl:176:                <?= eL('hours') ?>
CleanFS/templates/details.tabs.remind.tpl:29:          $how_often = $row['how_often'] / 3600 . ' ' . L('hours');
CleanFS/templates/details.tabs.remind.tpl:63:        <?php echo tpl_options(array(3600 => L('hours'), 86400 => L('days'), 604800 => L('weeks')), Req::val('timetype1')); ?>
CleanFS/templates/details.edit.tpl:128:		<?= eL('hours') ?>
```

trebmuh: No, it's not singular/plural but masculine/feminine. Next to this translation string, there is a menu choice between "Hours" (feminine), "Days" (masculine), and "Week" (feminine).

In French, it should then be "Toutes les heures", "Tous les jours", or "Toutes les semaines".
The issue with "Période" is that doesn't make sens as much as "Tou(te)s les". As an example, didn't had any clue of what this option was meaning/doing before reading the whole manual today. Hence this proposition.

trebmuh: No, it's not singular/plural but masculine/feminine. Next to this translation string, there is a menu choice between "Hours" (feminine), "Days" (masculine), and "Weeks" (feminine).

In French, it should then be "Toutes les heures", "Tous les jours", or "Toutes les semaines".
The issue with "Période" is that doesn't make sense as much as "Tou(te)s les" does. As an example, I didn't have any clue of what this option was meaning/doing before reading the whole manual today. Hence this proposition.

peterdd: ok.

The 'thisoften' keyword is IMHO ambiguous too, I stumbled over it too. Should better something like 'remindfrequency'.
In contrast to what I thought first like a possible max repeat count (not implemented) or max reminder end date (not implemented)

trebmuh: Thanks for the merge. :+1:

PR #755, opened 2020-09-10
Show subtasks effort in detail view

poucz:


peterdd:
Could you please make 2 separate pull requests? 

The proxy stuff is a separate issue I would like to be handled separately (there are similiar requests on bugs.flyspray.org)

includes/class.flyspray.php
```diff
@@ -390,6 +395,11 @@ public static function getTaskDetails($task_id, $cache_enabled = false)
             $get_details['assigned_to_name'] = $assignees[1];
         }
 	
+		global $proj;
+		if($proj->prefs['use_effort_tracking']){
```
peterdd, line 399: We cannot use the global $proj in function getTaskDetails(). It is necessary to use the project_id of task and then check if that project has use_effort_tracking set.

Neustradamus:
@poucz: Have you progressed on this PR?

PR #759, opened 2021-03-11
Fix FS#2628, access array offset on value of type int

HenryNe:
Fix FS#2628 - class.flyspray.php __construct(): Trying to access array offset on value of type int, PHP 7.4
for memory_limit = -1
The check "$val < 0" will skip negav´tive values also. If no meory_limt set, then one of the others will limit the size. Not need extra PHP_INT_MAX.


HenryNe:
PR #758 is better, because it will also catch settings like memory_limit = 100000000 in php.ini.

PR #775, opened 2021-08-04
Add sensible default for conf.php in .htaccess template

blu-base:
As far as i can tell, `/setup/index.php` creates the `flyspray.conf.php` in the root directory.

It is likely that new users will not modify the code to store `flyspray.conf.php` in an inherently secure location - out of the web root. That's why i think, that it is a good idea to provide the `.htaccess` which restricts access to the `conf.php`.

### DIFF
--- a/htaccess.dist
+++ b/htaccess.dist
@@ -55,6 +55,13 @@ php_flag always_populate_raw_post_data Off
 	deny from all
 </Files>
 
+# hide default location of the config
+<Files ~ "^flyspray.conf.php$">
+	Order allow,deny
+	Deny from all
+</Files>
+
+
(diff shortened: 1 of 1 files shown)


peterdd:
Mmh, the flyspray.conf.php itself has a check in the first line. So even if .htaccess is not enabled by webserver config it cannot be read from outside.

The improvement would be to get a 403 HTTP error instead of spitting `; Do not access this page directly.`, right?

blu-base:
It seems i did not think through the implications of the php fence line.

> The improvement would be to get a 403 HTTP error instead of spitting ; Do not access this page directly., right?

Yes you are right. It would become a second line of defense.

peterdd:
I think we do not need this.

PR #789, opened 2021-12-18
add compat for reverse proxy

kapouik:
Add compatibility for putting flyspray behind a reverse proxy.

### DIFF
--- a/includes/class.flyspray.php
+++ b/includes/class.flyspray.php
@@ -242,13 +242,19 @@ class Flyspray
         }
 
         if (empty($protocol)) {
-            if (isset($_SERVER['HTTPS']) && !strcasecmp($_SERVER['HTTPS'], 'on')) {
+            if (isset($_SERVER['HTTP_X_FORWARDED_PROTO']) && !strcasecmp($_SERVER['HTTP_X_FORWARDED_PROTO'], 'https')) {
+                $protocol = 'https';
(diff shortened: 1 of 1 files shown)


Neustradamus:
@peterdd: Maybe good to look this PR too?

peterdd:
The problem I have with this: I do not want merge stuff I am not yet understand/reproduce. I use flyspray in a hosting and seems to works fine with TLS, although it is behind a reverse proxy. So why it is needed then?

An example setup would be great, ideally provided as a Vagrantfile.

kapouik:
When you use a reverse proxy, there is 3 way to do the connexion : 

- Reverse proxy > https connexion > webserver : $_SERVER['HTTPS'] == on
- Reverse proxy > http connexion > webserver : $_SERVER['HTTPS'] == off
- Reverse proxy > proxy protocol > webserver : $_SERVER['HTTPS'] == http or https (depend what is use in front of the reverse proxy)

$_SERVER['HTTPS'] is normally set by the connexion made between the reverse proxy and the web server. There is standards variables used to know what is really used in front of the reverse proxy : x-forwarded-proto (you have x-forwarded-for to get the real IP of your visitor and x-forwarded-port to get the real port used). 

My patch add the possibility to get protocol and port from x-forwarded-proto and x-forwarded-port if they are set. If not, it will get the value from standard variables.

I don't use vagrant so I can't give you any conf for it.

Neustradamus:
@aronk02: I think that this PR is linked to your ticket:
- https://github.com/flyspray/flyspray/issues/935

PR #797, opened 2023-02-28
(Fix) set cache directory to prevent warnings

ms-syscon:
When the "SerializerPath" is not set, it will write into the "vendor" folder. This is critcal and not allowed on all servers. Why not use the existing "cache" folder?

### DIFF
--- a/includes/class.backend.php
+++ b/includes/class.backend.php
@@ -311,6 +311,7 @@ abstract class Backend
 
 	if ($conf['general']['syntax_plugin'] != 'dokuwiki') {
 		$purifierconfig = HTMLPurifier_Config::createDefault();
+    $purifierconfig->set('Cache.SerializerPath', 'cache');
 		$purifierconfig->set('CSS.AllowedProperties', array());
 		if ($fs->prefs['relnofollow']) {
 			$purifierconfig->set('HTML.Nofollow', true);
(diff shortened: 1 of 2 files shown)


peterdd:
I was not aware that HTMLpurifier write cache files to the file system and by default into the vendor/ dirs.

There is already a setting for cache directories in Flyspray - FS_CACHE_DIR

```php
grep -rn FS_CACHE_DIR
plugins/dokuwiki/dokuwiki_formattext.inc.php:59:		$conf['cachedir'] = FS_CACHE_DIR; // for dokuwiki
feed.php:50:$cachefile = sprintf('%s/%s', FS_CACHE_DIR, $filename);
includes/constants.inc.php:85:define('FS_CACHE_DIR', Flyspray::get_tmp_dir() . DIRECTORY_SEPARATOR . FS_DOMAIN_HASH);
includes/constants.inc.php:87:is_dir(FS_CACHE_DIR) || @mkdir(FS_CACHE_DIR, 0700);
```

Maybe this can be used too?

ms-syscon:
Thanks for the feedback. I've changed the code with FS_CACHE_DIR

Neustradamus:
@ms-syscon: Can you squash your commits to have only one?

@peterdd: Have you seen?

PR #798, opened 2023-03-14
Features flyspray/subtasks

ms-syscon:
- added sorting of subtasks by the header
- added an overview of the statuses above the subtasks

grafik

There are more things to do, but I will wait for your feedback. It would be great, if the subtasks have the same columns like the tasks in the project. The sorting should be work on all colums - like in the task list. And why not use the same presorting?

### DIFF
--- a/includes/modify.inc.php
+++ b/includes/modify.inc.php
@@ -478,17 +478,17 @@ switch ($action = Req::val('action'))
 				}
 			}
 		}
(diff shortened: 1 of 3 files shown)


peterdd (commented):
Please only change the relevant sections, no whitespace changes. Makes review much easier.

I am not sure if keeping the suborder in the querystring too is good idea..

Issue #842, opened 2023-12-14
Profile Image is very very small


Neustradamus:
When we upload a "Profile Image", it is very very small.

peterdd:
There is a setting for that in admin preferences:

![flyspray_profileimage_maxpixel](https://github.com/flyspray/flyspray/assets/1839154/3d18da8d-cfe0-4ef8-9986-d22f2994ae65)

The default value after a Flyspray install is 50 (Pixel).

It may be that in older installations this setting not existed and was not set to that default 50 value during upgrade and there is now a value 0 there. In that case uploaded profile images are resized and stored as only tiny 5x5 pixel sizes. :-/

As workaround set this value to 50.

Neustradamus:
Can you add a script in upgrade feature to all?

CaraesNaur:
Just to confirm @peterdd's theory...

I have an install I've been using since `1.0-rc1` which has been upgraded through:

* 1.0-rc4
* 1.0-rc7
* 1.0-rc10
* 1.0-rc11dev (current)

The `prefs` table has `max_avatar_size` set to `0`, so this seems not to have been set correctly during the relevant upgrades.

I am the only user on this install.  I have an avatar (48x48), mainly to show something in task comments.

While I was messing with my theme recently, I noticed the `max_avatar_size` was `0`, and looked into `tpl_userlinkavatar()`.  Multiple calls to that function are made by passing `$fs-prefs['max_avatar_size'] / 2` as `$size`, which results in a rather non-sensical `$avacache[$uid][0]` entry.

This may cause unexpected behavior with Gravatars where the image URI may specify size via `&s=0`.

I have modified my install to not use the `$size` argument to `tpl_userlinkavatar()`, my theme instead sizes avatars by CSS (I added a few avatar display during my most recent theme changes, such as in `Assigned To` in task details).

peterdd:
IMHO the main reason of setting max_avatar_size is for resizing images when people upload their profile image.

There are several places in the UI where a profile image can be shown. The size shown to user might differ set by CSS and should depend also on the screen size. For instance in task history the avatars are currently too big for my taste. Only little is needed there (~1 text line height) so the history stays dense 1 line per change entry.

As I am not using Gravatar, what is the 'unexpected behavior'?

Neustradamus:
An update of the SQL database will be the solution?
Can you add it?

CaraesNaur:
[...]

PR #883, opened 2024-06-10
Flyspray/Flyspray #842

CaraesNaur:
In includes/class.tpl.php:

Add protected class array $_avatar_sizes in FSTpl.

Add method FSTpl::get_avatar_sizes() which serves as a getter for this array.

Add method FSTpl::get_nearest_avatar_size($size) which returns one of three values from $avatar_sizes based on $size:

- If $size is less than least $avatar_sizes member, returns that member
- If $size is greater than greatest $avatar_sizes member, returns that member
- If $size is not in $avatar_sizes, returns the bext greater member

Added function tpl_avatar_sizes($selected), to render FSTpl::_avatar_sizes as an HTML <option> list.

In includes/modify.inc.php:

Within case 'globaloptions', use the newly added FSTpl class methods to validate the 'max_avatar_size' global preference.

In setup/upgrade.php:

Add $uplog to globals in execute_upgrade_file().

After deleting unwanted prefs, fetch all prefs in the upgrade version and loop over them to validate.  This is meant to catch & correct `max_avatar_size = 0` lingering after previous upgrades.

Added "if" block in the new prefs loop specifically to resolve this instance.

This loop appends entries to $uplog so that its effects can be displayed to the user.

In setup/upgrade/1.0/flyspray-install.xml:

Change default for 'max_avatar_size' to 320.

In setup/upgrade/1.0/upgrade.info:

Change 'max_avatar_size' to 320.

In themes/CleanFS/templates/admin.prefs.tpl:

Convert 'max_avatar_size' input to a <select> populated by tpl_avatar_sizes().


CaraesNaur:
@peterdd I expect you may want to modify the sizes list (it is a bit on the extensive side).

I wasn't sure where to house the list in the code, this was the cleanest solution I could come up with.

Using 320 as the default is rather arbitrary, but I think the default should be near the middle of the list as implemented.

Hmm... I forgot to fix all the instances where max_avatar_size is passed into `tpl_userlinkavatar()`.

peterdd:
Hmm, I don't think we need forcing a set of predefined sizes. Lets keep it simple.

peterdd:
And I think setup/upgrade.php should be quite generic. (former devs called it a mess and I touched only minimal/necessary parts, dont want it more messier)

Fiddling with a single upgrade issue in this file should be avoidedif possible. With entries in setup/upgrade/1.*/upgrade.info can PHP or xmlschema03 files be executed.

See `setup/upgrade/1.0/upgrade.info for example.

CaraesNaur:
I'll admit this is a bit of a long way round to solve the original issue.  However, a side effect of this approach is that `max_avatar_size` is now validated.

I was just able to store a non-digit string in this pref on my local Flyspray/Flyspray workspace.

Neustradamus:
To follow this PR

CaraesNaur:
Any input, @Neustradamus?

